PR #679, opened 2012-12-07
Bug 1065239 Add comment for 'unbind' function

ejpbruel:
The function 'unbind' was confusing for me at first. It's not obvious from the definition what it is supposed to do, so I've added a comment that describes it use.


erikvold:
Please re-open and rebase if/when you get a chance!

PR #917, opened 2013-04-03
Bug 847309 - Better handle widget user customization.

ochameau:
https://bugzilla.mozilla.org/show_bug.cgi?id=847309

Work in progress, need more testing with custom firefox builds.


erikvold:
we are removing widget module in [bug 1118364](https://bugzilla.mozilla.org/show_bug.cgi?id=1118364)

PR #938, opened 2013-04-13
Bug 860419 - Let users choose whether addon supports private browsing.

Gozala:
WIP -

This change is going add a preference to every SDK add-on allowing users to disable add-on for private browsing.
If preference is set to `false` it will make [self.isPrivateBrowsingSupported](https://github.com/mozilla/addon-sdk/blob/master/lib/sdk/self.js#L40-L41) to be `false` hiding all the private browsing windows from the high level APIs.
Switching preference will undo all the changes that add-on did to a private windows (Initially probably we could just disable and re-enable addon with new pref set. We can optimize it later).


andymckay:
Closing due to inactivity.

PR #947, opened 2013-04-16
Only emit to non-null _contentWorker

jxpx777:
This prevents logging exception data that can easily be avoided. Some of our users notice these and complain about them cluttering their consoles.


andymckay:
Closing due to inactivity.

PR #949, opened 2013-04-17
Bug 854980 - Tab detraitification.

Gozala:


erikvold:
tabs have been detraitified.

PR #1059, opened 2013-06-27
Bug 820953 - Expose web-socket API.

Gozala:
This pull was intentionally separated out of #1049 so that node style net module and this can land separately.


andymckay:
Bug 820953 is marked as resolved fixed in bugzilla.

PR #1125, opened 2013-07-18
Bug 565388 - Switch to using permanent windowless docshells.

Gozala:
Work in progress switch to permanent docshells.


andymckay:
Bug 565388 is resolved fixed, so marking as closed.

PR #1133, opened 2013-07-24
Bug 897370 - Use mozrunner from mozbase upstream

nodeless:


andymckay:
Closed due to inactivity, the bug in bugzilla was also marked as resolved wont fix.

PR #1220, opened 2013-08-24
+sdk/io/file method (reveal)

brettz9:


brettz9:
Terribly sorry not to have replied to this earlier. I was having trouble setting up an environment to do testing, and I am putting my limited energies into other work at the moment.

PR #1279, opened 2013-11-04
Bug 934696 Node Loader JEP

jsantell:
The loader change would be a change to how files are resolved when using Jetpack's require to accomodate node modules.

[Meta Bug 934696](https://bugzilla.mozilla.org/show_bug.cgi?id=934696)


andymckay:
Closing due to inactivity.

PR #1319, opened 2013-12-07
Bug 795522 A basic declarative tab context

sachinhosmani:
Since "tabContextMenu" is the ID of the browser tab context menu, the new items that need to get added go right there.


andymckay:
Closing due to inactivity.

PR #1331, opened 2013-12-21
Bug 952265 - Expose tab 'rename' and 'move' events.

RyanStein:
https://bugzilla.mozilla.org/show_bug.cgi?id=952265

This commit implements a 'rename' event and exposes the 'move' event
which was previously hidden. The code for Fennec has yet to be tested.


andymckay:
Going to close based on inactivity, if this is going to continue we can do so in the bugzilla bug. Sounds like Irakli volunteered to take it over in this bug https://bugzilla.mozilla.org/show_bug.cgi?id=952265#c8.

PR #1335, opened 2014-01-08
Bug 935120 Node events module

jsantell:
* Moved `sdk/event/target` to `node/events` -- changed exposed object from `EventTarget` to `EventEmitter`. No deprecation until we're ready.
* Added `removeAllListeners(type)` to EventEmitter. Similar to off(target) and off(target, type).
* Added `listeners(type)` to return listener count
* Added `emitter.emit` return boolean indicating if there were any listeners listening
* Added static method `EventEmitter.listenerCount()`, same as listeners(type)
* Added events 'newListener' and 'removeListener' to be fired upon adding/removing listeners.
* Added alias for `on`: `addListener`
* Node's `setMaxListeners()` was noop'd, but not implemented here -- may incur too much overhead.


andymckay:

PR #1336, opened 2014-01-09
Bug 935122 - Move querystring over to node API

jsantell:
Leaving deprecation of sdk/querystring for a future step once we have native loader


andymckay:

PR #1349, opened 2014-01-15
Bug 889218: add tests for 'escape' closing panel when open

jsantell:


andymckay:

PR #1412, opened 2014-02-26
Work in progress implementation for devtools pane API

Gozala:


andymckay:
Going to close based on inactivity being a work in progress bug.

PR #1486, opened 2014-05-11
bug 788224, part 1, added url and contentType to include

darkowlzz:


andymckay:

PR #1502, opened 2014-05-31
Bug 980555: Created JEP for Object Inspectors.

cbrem:


andymckay:

PR #1518, opened 2014-06-17
Bug 1026796 - Allow tags, annotation, and frecency sorting

brettz9:
Not tested but appears to match values at https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsINavHistoryQueryOptions#Sorting_methods 

(I'm currently only interested in frecency (for applying the awesomebar approach to my `file://` browser overlay add-on), but thought all of these ought to be included).


andymckay:

PR #1597, opened 2014-08-20
Bug 661083 manifest localization

erikvold:


anaran:
How would this kind of localization work with jpm tooling?

PR #1615, opened 2014-09-09
Devtools SDK - Director JEP draft

rpl:
initial JEP draft for the Director, a proposed Devtools SDK component which helps Devtools Addon developers to inject their instrumentation code into the debugging target (e.g. a tab or an app).


andymckay:
Assuming this isn't happening, but I'm sure @rpl will tell me if I'm wrong.

PR #1669, opened 2014-10-01
Bug 1036625: Loader shouldn't be attempting to load files just to resolve modules.

Mossop:
This corrects the minor typo from the previous patch and adds a test that uses the native loader with a packed add-on to verify that it can resolve both normal and node modules either as plain js, directories with index.js or directories with main mappings.


andymckay:
Closed due to inactivity and the bug not being clear on the resolution at this point.

PR #1670, opened 2014-10-02
Bug 889465 - Buttons need GC tests

ZER0:
This PR is made on top of the badge's one, because it was already r+ and improves the button's tests.
But is not landed yet because it depends by a platform one, so the only commit to be review is the following: https://github.com/ZER0/addon-sdk/commit/f71e768992a537b4839c687173d13b740d7b1296


ZER0:
fix Bug 889465 - Buttons need GC tests r=@erikvold

ZER0:
@Xiller you need also a patch in Firefox, so it could be that you tried this code with a Firefox version that doesn't support badged button yet. The badged button now is landed since a while, so it shouldn't be any issue I hope!

PR #1700, opened 2014-11-05
Bug 1094392: Fix weak referenced observers

Mossop:


Mossop:
This testcase definitely still fails and from what I can see the bug it is testing for is what is causing most of the intermittent oranges in test-panel.

Mossop:
The root cause is a platform bug which I've filed bug 1131317 for. Once that is fixed we should probably take this test for it.

test/test-observers.js
```diff
@@ -120,4 +120,43 @@ exports["test weak observers are GC-ed on unload"] = (assert, end) => {
...
+  observe.define(WeakObserver, (x, ...rest) => x.observe(...rest));
+
+  let observer = new WeakObserver;
+  subscribe(observer, "test-topic");
+
+  notifyObservers(null, "test-topic", null);
+  assert.ok(sawNotification, "Should have seen notification before GC");
+  sawNotification = false;
+
+  loader.require("sdk/test/memory").gc();
```
erikvold: yield this line

test/test-observers.js
```diff
@@ -120,4 +120,43 @@ exports["test weak observers are GC-ed on unload"] = (assert, end) => {
   });
 };
 
-require("sdk/test").run(exports);
\ No newline at end of file
+exports["test weak observer unsubscribe"] = (assert, end) => {
```
erikvold: let's make this one a generator

test/test-observers.js
```diff
@@ -120,4 +120,43 @@ exports["test weak observers are GC-ed on unload"] = (assert, end) => {
...
+    observe: function() {
+      sawNotification = true;
+    }
+  });
+  observe.define(WeakObserver, (x, ...rest) => x.observe(...rest));
+
+  let observer = new WeakObserver;
+  subscribe(observer, "test-topic");
+
+  notifyObservers(null, "test-topic", null);
```
erikvold: we should yield on promise from the observer here.

test/test-observers.js
```diff
@@ -120,4 +120,43 @@ exports["test weak observers are GC-ed on unload"] = (assert, end) => {
...
+  assert.ok(sawNotification, "Should have seen notification before GC");
+  sawNotification = false;
+
+  loader.require("sdk/test/memory").gc();
+
+  notifyObservers(null, "test-topic", null);
```
erikvold: we should yield on promise from another observer here either for the same topic or another one, to keep the test async.

test/test-observers.js
```diff
@@ -120,4 +120,43 @@ exports["test weak observers are GC-ed on unload"] = (assert, end) => {
...
+  assert.ok(sawNotification, "Should have seen notification after GC");
+  sawNotification = false;
+
+  try {
+    unsubscribe(observer, "test-topic");
+    assert.ok(true, "Should not have seen an exception");
+  }
+  catch (e) {
+    assert.ok(false, "Should not have seen an exception");
```
erikvold: nit use `assert.fail` here

test/test-observers.js
```diff
@@ -120,4 +120,43 @@ exports["test weak observers are GC-ed on unload"] = (assert, end) => {
...
+
+  loader.require("sdk/test/memory").gc();
+
+  notifyObservers(null, "test-topic", null);
+  assert.ok(sawNotification, "Should have seen notification after GC");
+  sawNotification = false;
+
+  try {
+    unsubscribe(observer, "test-topic");
+    assert.ok(true, "Should not have seen an exception");
```
erikvold: nit use `assert.pass` here

test/test-observers.js
```diff
@@ -120,4 +120,43 @@ exports["test weak observers are GC-ed on unload"] = (assert, end) => {
...
+  }
+  catch (e) {
+    assert.ok(false, "Should not have seen an exception");
+  }
+  end();
```
erikvold: `loader.unload()` here

lib/sdk/core/observer.js
```diff
@@ -41,13 +41,12 @@ const ObserverDelegee = Class({
     this.delegate = delegate;
   },
   QueryInterface: function(iid) {
-    const isObserver = iid.equals(Ci.nsIObserver);
-    const isWeakReference = iid.equals(Ci.nsISupportsWeakReference);
+    if (iid.equals(Ci.nsIObserver) ||
+        iid.equals(Ci.nsISupportsWeakReference) ||
+        iid.equals(Ci.nsISupports))
+      return this;
```
erikvold: I think this organization will result in a strict warning about this function not returning.  Can we keep the old `if` `throw` `return` organization?

test/test-observers.js
Mossop: I don't understand what that adds to the test. The observer service is
synchronous, it guarantees that all observers will have been called when
notifyObservers returns, adding a second observer seems like a no-no.

lib/sdk/core/observer.js
Mossop: We can do but I suspect it's not a problem, this is basically the standard form for QueryInterface: http://mxr.mozilla.org/mozilla-central/source/js/xpconnect/loader/XPCOMUtils.jsm#339

erikvold: afacit that file isn't using strict mode.

test/test-observers.js
erikvold: this test isn't using the observer service directly though, I'm not sure we want to guarantee `sdk/core/observer` is synchronous.

Mossop:
Landed in 0aa270ada80c70c1690c9befedf917eca7555d7b

PR #1726, opened 2014-11-20
Bug 724632 - Detraitify

Gozala:
# [Bug 724632](https://bugzilla.mozilla.org/show_bug.cgi?id=sdk-kill-traits) Remove use of traits

This is another attempt to kill all traits with fire! Although this time I'm less ambitious and just trying to replace traits with Class and private property trickery with symbols. Unfortunately it's nearly impossible to land this change in chunks because there are few common traits shared between many APIs. Since it's also would be inhumane to review the whole thing (this pull) I'm going to use following strategy. Create pull request for each small component that has tests that pass. We won't be able to land those pulls separately as that would break users of that component, but this way we could verify review and verify each peace & and aggregate all changes in this pull. Once all pieces are r+ and all tests pass in this pull we can land it.

Still there is still at least one piece that may land separately so I'm going to create two sections here where I'll link changes for the pieces this pull request contains.

## Standalone things that can land

- [x] #1733
- [x] #1724 

## Pieces that are coupled

- [ ] #1725
- [ ] #1734 (implicitly depends on #1733 & #1725)

### DIFF
--- a/lib/sdk/content/loader.js
+++ b/lib/sdk/content/loader.js
@@ -8,7 +8,9 @@ module.metadata = {
   "stability": "unstable"
 };
 
-const { EventEmitter } = require('../deprecated/events');
+const { Class } = require('../core/heritage');
+const { EventTarget } = require('../event/target');
+const { emit } = require('../event/core');
 const { isValidURI, isLocalURL, URL } = require('../url');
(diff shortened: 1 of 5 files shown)


erikvold:
Traits have been removed

PR #1728, opened 2014-11-21
Bug 1100644: Use browser.hiddenWindowChromeURL as a fallback since brows...

fred-wang:
...er.ChromeURL is not defined on all Thunderbird versions. r=erikvold


fred-wang:
abandoned

PR #1735, opened 2014-11-25
Bug 1057346 - Devtools styling API

janodvarko:


ZER0:
I apologize for the delay! Looks good to me, r+. However I'd like to have some unit test to be sure that the API is working as expected; especially for the methods `getStyles` and `getClassList` when we pass strings, arrays and themes. We could also implement the support for `Style` object but I don't think is a blocker, we can implement that in the future, especially if we're going to use more widely the `Style` object.

PR #1772, opened 2014-12-26
seems to fix problem where resource uris are generated incorrectly

canuckistani:
See [bug 1115601](https://bugzilla.mozilla.org/show_bug.cgi?id=1115601) for more info, this seems to break jpm resource uris.


canuckistani:
This is obsolete.

PR #1813, opened 2015-01-07
Bug 839983 Test that window.title is undefined after the window is closed

erikvold:


test/windows/test-firefox-windows.js
```diff
@@ -463,4 +463,29 @@ exports["test getView(window)"] = function(assert, done) {
...
+  assert.ok(sdkWindow.title, "title is not undefined");
+
+  yield focus(rawWindow);
+
+  assert.pass("the window was focused");
+
+  yield close(rawWindow);
+
+  assert.pass("the window was closed");
+  assert.equal(sdkWindow.title, undefined, "window.title is undefined");
```
zombie: nit: use `strictEqual()` to be explicit here

PR #1815, opened 2015-01-07
Bug 1118639 - Write before and after tests for sidebars to check that no sidebars are showing

erikvold:


erikvold:
Closing for #1818

PR #1826, opened 2015-01-08
bug 1060695 - add a frame script for selection events

zombie:


lib/framescript/selection-events.js
```diff
@@ -0,0 +1,45 @@
...
+function observer(subject, topic) {
+  if (!docShell)
+    obsSvc.removeObserver(observer, topic);
+  else {
+    let window = subject.defaultView;
+    if (!window || window.top !== content)
+      return;
+
+    let selection = window.getSelection();
+    if (!selection || selections.get(window) === selection)
```
Mossop: When is selection ever null?

lib/framescript/selection-events.js
```diff
@@ -0,0 +1,45 @@
...
+
+obsSvc.addObserver(observer, 'document-shown', false);
+obsSvc.addObserver(observer, 'document-element-inserted', false);
+
+function observer(subject, topic) {
+  if (!docShell)
+    obsSvc.removeObserver(observer, topic);
+  else {
+    let window = subject.defaultView;
+    if (!window || window.top !== content)
```
Mossop: When is window ever null?

Mossop: Please add a comment to the effect that we get observer notifications for every frame in the process so we're filtering to only the ones for this frame.

lib/framescript/selection-events.js
```diff
@@ -0,0 +1,45 @@
...
+
+const INTERACTIVE = Ci.nsISelectionListener.MOUSEUP_REASON |
+                    Ci.nsISelectionListener.KEYPRESS_REASON |
+                    Ci.nsISelectionListener.SELECTALL_REASON;
+
+let selections = new WeakMap();
+let isPrivate = docShell.QueryInterface(Ci.nsILoadContext).usePrivateBrowsing;
+
+obsSvc.addObserver(observer, 'document-shown', false);
+obsSvc.addObserver(observer, 'document-element-inserted', false);
```
Mossop: Might be worth testing if using the DOM event "DOMWindowCreated" is better than using global observer notifications.

zombie (commented):
Looks great, thanks!

 [blind-review](https://github.com/zombie/blind-reviews 'Review performed in blind mode')

PR #1829, opened 2015-01-09
Bug 1075490 - Fix devtools docking.

Gozala:
Depends on platform change attached here (https://bug1075490.bugzilla.mozilla.org/attachment.cgi?id=8546311)

### DIFF
--- a/lib/dev/frame-script.js
+++ b/lib/dev/frame-script.js
@@ -72,6 +72,9 @@ addMessageListener("sdk/event/message", onInMessage);
 addMessageListener("sdk/port/message", onInPort);
 
 const observer = {
+  handleEvent: ({target, type}) => {
+    this.observe(content.document, type);
+  },
   observe: (document, topic, data) => {
     // When frame associated with message manager is removed from document `docShell`
     // is set to `null` but observer is still kept alive. At this point accesing
@@ -96,7 +99,7 @@ const observer = {
           uri: document.documentURI
         });
       }
-      if (topic === "content-page-hidden") {
+      if (topic === "unload") {
         channels.clear();
         sendAsyncMessage("sdk/event/unload", {
           type: "unload",
@@ -110,6 +113,6 @@ const observer = {
 
 observerService.addObserver(observer, "content-document-interactive", false);
 observerService.addObserver(observer, "content-document-loaded", false);
-observerService.addObserver(observer, "content-page-hidden", false);
+addEventListener("unload", observer, false);
 
 })(this);
--- a/lib/dev/frame-script.js
+++ b/lib/dev/frame-script.js
@@ -73,7 +73,7 @@ addMessageListener("sdk/port/message", onInPort);
 
 const observer = {
   handleEvent: ({target, type}) => {
-    this.observe(content.document, type);
+    observer.observe(content.document, type);
   },
   observe: (document, topic, data) => {
     // When frame associated with message manager is removed from document `docShell`
@@ -85,14 +85,14 @@ const observer = {
       observerService.removeObserver(observer, topic);
     }
     else if (document === content.document) {
-      if (topic === "content-document-interactive") {
+      if (topic.endsWith("-document-interactive")) {
         sendAsyncMessage("sdk/event/ready", {
           type: "ready",
           readyState: document.readyState,
           uri: document.documentURI
         });
       }
-      if (topic === "content-document-loaded") {
+      if (topic.endsWith("-document-loaded")) {
         sendAsyncMessage("sdk/event/load", {
           type: "load",
           readyState: document.readyState,
@@ -113,6 +113,8 @@ const observer = {
 
 observerService.addObserver(observer, "content-document-interactive", false);
 observerService.addObserver(observer, "content-document-loaded", false);
+observerService.addObserver(observer, "chrome-document-interactive", false);
+observerService.addObserver(observer, "chrome-document-loaded", false);
 addEventListener("unload", observer, false);
 
 })(this);
--- a/lib/dev/frame-script.js
+++ b/lib/dev/frame-script.js
@@ -73,7 +73,7 @@ addMessageListener("sdk/port/message", onInPort);
 
 const observer = {
   handleEvent: ({target, type}) => {
-    observer.observe(content.document, type);
+    observer.observe(target, type);
   },
   observe: (document, topic, data) => {
     // When frame associated with message manager is removed from document `docShell`
--- a/lib/dev/panel.js
+++ b/lib/dev/panel.js
@@ -230,8 +230,6 @@ createView.define(Panel, (panel, document) => {
     // We end up using chrome iframe with forced message manager
     // as fixing a swapFrameLoader seemd like a giant task (see
     // Bug 1075490).
-    // "remote": true,
-    // "type": "content",
     "type": "chrome",
     "forcemessagemanager": true,
     "transparent": true,
--- a/lib/dev/panel.js
+++ b/lib/dev/panel.js
@@ -227,11 +227,13 @@ createView.define(Panel, (panel, document) => {
   const frame = document.createElement("iframe");
   setAttributes(frame, {
     "sandbox": "allow-scripts",
-    // It would be great if we could allow remote iframes for sandboxing
-    // panel documents in a content process, but for now platform implementation
-    // is buggy on linux so this is disabled.
+    // We end up using chrome iframe with forced message manager
+    // as fixing a swapFrameLoader seemd like a giant task (see
+    // Bug 1075490).
     // "remote": true,
-    "type": "content",
+    // "type": "content",
+    "type": "chrome",
+    "forcemessagemanager": true,
     "transparent": true,
     "seamless": "seamless",
   });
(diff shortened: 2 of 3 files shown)


lib/dev/panel.js
```diff
@@ -227,11 +227,13 @@ createView.define(Panel, (panel, document) => {
   const frame = document.createElement("iframe");
   setAttributes(frame, {
     "sandbox": "allow-scripts",
-    // It would be great if we could allow remote iframes for sandboxing
-    // panel documents in a content process, but for now platform implementation
-    // is buggy on linux so this is disabled.
+    // We end up using chrome iframe with forced message manager
+    // as fixing a swapFrameLoader seemd like a giant task (see
```
jryans: Nit: seemd -> seemed

lib/dev/panel.js
```diff
@@ -227,11 +227,13 @@ createView.define(Panel, (panel, document) => {
...
+    // Bug 1075490).
     // "remote": true,
-    "type": "content",
+    // "type": "content",
```
erikvold: nit remove this comment

lib/dev/frame-script.js
```diff
@@ -72,6 +72,9 @@ addMessageListener("sdk/event/message", onInMessage);
 addMessageListener("sdk/port/message", onInPort);
 
 const observer = {
+  handleEvent: ({target, type}) => {
+    this.observe(content.document, type);
```
erikvold: I'm guessing that `this.observe` is undefined here? @gozala does that make sense? and we should have a test that this works.

lib/dev/frame-script.js
```diff
@@ -110,6 +113,6 @@ const observer = {
 
 observerService.addObserver(observer, "content-document-interactive", false);
 observerService.addObserver(observer, "content-document-loaded", false);
-observerService.addObserver(observer, "content-page-hidden", false);
+addEventListener("unload", observer, false);
```
erikvold: where is this defined?

lib/dev/panel.js
```diff
@@ -227,11 +227,13 @@ createView.define(Panel, (panel, document) => {
...
+    "type": "chrome",
```
erikvold: Do we have a test that remote content is not used?

lib/dev/frame-script.js
Gozala: Scope in which frame scripts execute have `addEventListener` defined.

lib/dev/panel.js
Gozala: > Do we have a test that remote content is not used?

I don't think that's relevant to this change, but yes we options are validated:

https://github.com/mozilla/addon-sdk/blob/master/lib/dev/panel.js#L158-L164

lib/dev/frame-script.js
Gozala: > I'm guessing that this.observe is undefined here? @gozala does that make sense? and we should have a test that this works.

You are right! I had method before just changed it to arrow to match `observe` & have not run tests again. I'll update that.

Yes we do have tests for unload that were passing until I've changed method to arrow function.

test/test-dev-panel.js
```diff
@@ -150,6 +169,7 @@ exports["test Panel communication"] = test(function*(assert) {
   yield closeToolbox();
 
   assert.equal(panel.readyState, "destroyed", "panel is destroyed");
+  myTool.destroy();
```
erikvold: this seems unrelated, if it is then please make a separate bug for this change.

test/test-dev-panel.js
```diff
@@ -86,6 +86,25 @@ exports["test Panel API"] = test(function*(assert) {
   yield closeToolbox();
 
   assert.equal(panel.readyState, "destroyed", "panel is destroyed");
+
+  myTool.destroy();
+});
+
+
```
erikvold: we should add a comment mentioning that this is important since we use a "chrome" frame, and include a link to this bug too.

test/test-dev-panel.js
```diff
@@ -86,6 +86,25 @@ exports["test Panel API"] = test(function*(assert) {
...
+exports["test forbid remote docs"] = test(function*(assert) {
+  const MyPanel = Class({
+    extends: Panel,
+    label: "test panel",
+    tooltip: "my test panel",
+    icon: iconURI,
+    url: "https://mozilla.org",
+  });
+
```
erikvold: another test for `http` too please.

erikvold: another test for `ftp` too please.

erikvold:
We should have a test that it's not possible to have the local content (which passes validation) redirect to non-local content.

erikvold:
This is failing on travis:

7114 of 7118 tests passed.

JPM info There were test failures...
JPM info 
The following tests failed:

JPM info   ./test/test-dev-panel.test Panel API: timed out
  ./test/test-dev-panel.test Panel communication: timed out

JPM info   ./test/test-dev-panel.test communication with debuggee: timed out
  ./test/test-dev-panel.test viewFor panel: timed out

erikvold:
>> I'm guessing that this.observe is undefined here? @gozala does that make sense? and we should have a test that this works.

> You are right! I had method before just changed it to arrow to match `observe` & have not run tests again. I'll update that.

> Yes we do have tests for unload that were passing until I've changed method to arrow function. 

Which ones?

lib/dev/frame-script.js
```diff
@@ -82,21 +85,21 @@ const observer = {
       observerService.removeObserver(observer, topic);
     }
     else if (document === content.document) {
-      if (topic === "content-document-interactive") {
+      if (topic.endsWith("-document-interactive")) {
```
erikvold: this is an unrelated change, please separate this change in to another bug.

lib/dev/frame-script.js
```diff
@@ -110,6 +113,8 @@ const observer = {
 
 observerService.addObserver(observer, "content-document-interactive", false);
 observerService.addObserver(observer, "content-document-loaded", false);
-observerService.addObserver(observer, "content-page-hidden", false);
+observerService.addObserver(observer, "chrome-document-interactive", false);
+observerService.addObserver(observer, "chrome-document-loaded", false);
```
erikvold: why do you need both chrome and content observers here?

lib/dev/frame-script.js
```diff
@@ -72,6 +72,9 @@ addMessageListener("sdk/event/message", onInMessage);
 addMessageListener("sdk/port/message", onInPort);
 
 const observer = {
+  handleEvent: ({target, type}) => {
+    observer.observe(content.document, type);
```
erikvold: can you use the `target` here please, or remove the argument variable.

lib/dev/frame-script.js
Gozala: > this is an unrelated change, please separate this change in to another bug.

It is related because if content is chrome privileged topic will be `chrome-document-interactive"

lib/dev/frame-script.js
Gozala: > @gozala why do you need both chrome and content observers here?

Because it could be chrome privileged document.

test/test-dev-panel.js
Gozala: > we should add a comment mentioning that this is important since we use a "chrome" frame, and include a link to this bug too.

I do not understand this. You want me to comment why I destroy a tool here ?

Gozala:
> This is failing on travis:
> 
> 7114 of 7118 tests passed.
> 
> JPM info There were test failures...
> JPM info
> The following tests failed:
> 
> JPM info ./test/test-dev-panel.test Panel API: timed out
> ./test/test-dev-panel.test Panel communication: timed out
> 
> JPM info ./test/test-dev-panel.test communication with debuggee: timed out
> ./test/test-dev-panel.test viewFor panel: timed out

I've already pointed out that travis will fail .as it depend on platform change

Gozala:
> We should have a test that it's not possible to have the local content (which passes validation) redirect to non-local content.

I'm afraid I can't make such guarantees, since we make this local content chrome privileged it can do whatever it pleased no matter how hard we try.

Gozala:
@erikvold I removed those `myTool.destroy()` calls now as well. But honestly I think you are not consistent with your requests, in some cases you ask me to add unrelated tests and in other cases you ask me to remove changes that ensure proper cleanup after test run.

Also now that related platform changes are in nightly travis is happy and passing tests!

Gozala:
Created separate pull for test cleanups as requested here #1847

Gozala:
Also here is a link to unload tests that came up in the discussion #1848

lib/dev/frame-script.js
```diff
@@ -110,6 +113,8 @@ const observer = {
 
 observerService.addObserver(observer, "content-document-interactive", false);
 observerService.addObserver(observer, "content-document-loaded", false);
-observerService.addObserver(observer, "content-page-hidden", false);
+observerService.addObserver(observer, "chrome-document-interactive", false);
+observerService.addObserver(observer, "chrome-document-loaded", false);
+addEventListener("unload", observer, false);
```
ochameau: Shouldn't you remove this listener at some point? Same question for observerService?
Or is this frame script global and want to dispatch events for all document living in the tab?

lib/dev/frame-script.js
ochameau: It looks like test-dev-panel only has panels using data: URI.
It would be great if you could also test a chrome:// URL, like `chrome://global/content/about.xhtml` for ex, can we load any document or does it has to do something special to work in panel?

PR #1835, opened 2015-01-18
Bug 1122978 using the jsloader.reuseGlobal pref

erikvold:


erikvold:
this doesn't work unless the app is b2g.

PR #1838, opened 2015-01-21
Bug 1118364 - Remove widget module from SDK

ZER0:
- Widget module removed as well as the unit test
- Widget examples removed


test/addons/layout-change/lib/test-cuddlefish-loader.js
```diff
@@ -21,11 +21,6 @@ exports["test compatibility"] = function(assert) {
   assert.equal(require("tabs"),
                require("sdk/tabs"), "sdk/tabs -> tabs");
 
-  if (app.is("Firefox")) {
-    assert.equal(require("widget"),
-                 require("sdk/widget"), "sdk/widget -> widget");
-  }
```
erikvold: this should now test that the mapping does not exist, and we need to remove this mapping.

test/test-panel.js
```diff
@@ -893,34 +893,34 @@ exports['test passing DOM node as first argument'] = function (assert, done) {
...
+        'the panel is properly anchored to the button');
 
       shown.resolve();
     }
   });
 
-  let widget = Widget({
+  let button = ActionButton({
     id: 'panel-widget',
     label: 'panel widget',
```
erikvold: this id and label probably should be updated

test/addons/layout-change/lib/test-cuddlefish-loader.js
ZER0: I'm not sure why we should test for a module that it doesn't exist anymore, I'm missing the point to understand what's the advantage to do so. It seems to me that we're still adding a reference to widget module that soon or later we have to remove again.

test/test-panel.js
ZER0: "widget" is a generic term, is used also in firefox to identify buttons; that's why I didn't update it; but I do not have issue to change label and id, it's fine to me.

test/addons/layout-change/lib/test-cuddlefish-loader.js
erikvold: keeping the "widget" -> "sdk/widget" mapping would break a node_module named widget, and since we don't need it any longer we should remove.  We shouldn't keep code around that doesn't do anything right?

ZER0: No, that's clear. I forgot to remove "widget" from mapping, and I have to do that in my next commit. I was questioning the first part, that we should test that the mapping does not exist. I don't agree with that, because doing so we actually introduce a code that is referring to a module that doesn't exist anymore, an soon or later we have to remove again – and we probably forget to do so.
So my point was just to remove the reference in all our codes – including mapping and unit test, without adding a unit test check for a module that doesn't exist.

erikvold: @zer0 I think here or elsewhere we should test what happens with the cuddlefish loader when the widget module is required and make sure that what we want to see happen actually happens.  Since we know there will be plenty of people that hit this use case.

ZER0: @erikvold I'm not sure I'm following you. If we remove all the `widget` from SDK, even the mapping – that I've done now – what we should test? I mean, requiring the `widget` (or `sdk/widget`) module, with this patch, is basically like requiring a `foo` module: it doesn't exist. So what should be the purpose of the test? You're suggesting to test that every module we remove it's actually failing in requiring them? We never did that in the past, and I don't see the advantage, just adding reference to non-existing modules… and moreover, where we should add those tests?
If I didn't understand what you mean, please provide the code you have in mind. Thanks!

test/test-panel.js
erikvold: ok

test/addons/layout-change/lib/test-cuddlefish-loader.js
erikvold: > what we should test?

    // sdk/widget module was removed
    assert.throws(() => require("widget"), /.../, "There is no widget module");
    assert.throws(() => require("sdk/widget"), /.../, "There is no sdk/widget module");

> So what should be the purpose of the test? You're suggesting to test that every module we remove it's actually failing in requiring them?

Every high level module yes.

First of all we've never removed a high level module in the past, so this is a unique case.  I'm suggesting we should test that every high level module we remove is actually gone, so that we don't add it again either on purpose or by mistake, and cause regressions.

> And moreover, where we should add those tests?

I don't think this matters, we can make a new test file for it or use an old one.

PR #1840, opened 2015-01-22
Bug 1098391 - Best effort to load SDK loader dependencies lazily

ochameau:
https://bugzilla.mozilla.org/show_bug.cgi?id=1098391


lib/toolkit/loader.js
```diff
@@ -744,7 +738,15 @@ function Loader(options) {
...
     globals: {
-      console: console
+      get console() {
+        // Import Console.jsm from here to prevent loading it until someone uses it
+        let { ConsoleAPI } = Cu.import("resource://gre/modules/devtools/Console.jsm");
+        let console = new ConsoleAPI({
+          consoleID: options.id ? "addon/" + options.id : ""
+        });
+        Object.defineProperty(this, "console", { value: console });
+        return console;
```
Gozala: Can you please return `this.console` instead so we're sure it's end replacing a getter.

PR #1844, opened 2015-01-27
Bug 1012988 - Ensure that the loader doesn't force loader magic modules

ochameau:
https://bugzilla.mozilla.org/show_bug.cgi?id=1012988

### DIFF
--- a/lib/toolkit/loader.js
+++ b/lib/toolkit/loader.js
@@ -703,7 +703,8 @@ Loader.main = main;
 const Module = iced(function Module(id, uri) {
   return create(null, {
     id: { enumerable: true, value: id },
-    exports: { enumerable: true, writable: true, value: create(null) },
+    exports: { enumerable: true, writable: true, value: create(null),
+               configurable: true },
     uri: { value: uri }
   });
 });
@@ -784,6 +785,7 @@ function Loader(options) {
     }
   }, modules);
 
+  const builtinModuleExports = modules;
   modules = keys(modules).reduce(function(result, id) {
     // We resolve `uri` from `id` since modules are cached by `uri`.
     let uri = resolveURI(id, mapping);
@@ -792,7 +794,16 @@ function Loader(options) {
     if (isNative && !uri)
       uri = id;
     let module = Module(id, uri);
-    module.exports = freeze(modules[id]);
+
+    // Lazily expose built-in modules in order to
+    // allow them to be loaded lazily.
+    Object.defineProperty(module, "exports", {
+      enumerable: true,
+      get: function() {
+        return builtinModuleExports[id];
+      }
+    });
+
     result[uri] = freeze(module);
     return result;
   }, {});
--- a/test/test-loader.js
+++ b/test/test-loader.js
@@ -506,4 +506,22 @@ exports["test Cu.import in b2g style"] = (assert) => {
                "Loader.Loader is a funciton");
 };
 
+exports['test lazy globals'] = function (assert) {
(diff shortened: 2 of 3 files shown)


lib/toolkit/loader.js
```diff
@@ -785,7 +786,20 @@ function Loader(options) {
     if (isNative && !uri)
       uri = id;
     let module = Module(id, uri);
-    module.exports = freeze(modules[id]);
+
+    // Lazily expose built-in modules in order to
+    // allow them to be loaded lazily.
+    let cachedModules = modules;
```
Gozala: This code is very confusing! I had to open full file and it took me few passes to grok it. So if I do understand it correctly now you capture passed modules into `cachedModules` just before it `modules` is set to this new object in order to allow lazy access. If that so here are my suggestion:

No need to copy original `modules` in each iteration, specially in the callback since it's not obvious what `modules` is at this point. I would suggest to create `const builtinModuleExports = modules` on the line above `modules = ...`.

lib/toolkit/loader.js
```diff
@@ -785,7 +786,20 @@ function Loader(options) {
...
+    Object.defineProperty(module, "exports", {
+      enumerable: true,
+      get: function() {
+        return cachedModules[id];
+      },
+      set: function(v) {
```
Gozala: Is there a use case for the setter ? If so please add a comment explaining it. Although I really hope we don't need a setter here, in which case I'd suggest to either remove it or make one that throws exception explaining why this is forbidden.

test/test-loader.js
```diff
@@ -456,4 +456,22 @@ exports['test loader on unsupported modules with checkCompatibility default'] =
...
+      return foo;
+    }
+  };
+  let loader = Loader({ paths: { '': uri }, globals: globals});
+  assert.ok(!gotFoo, "foo hasn't been accessed during loader instanciation");
+  let program = main(loader, 'main');
+  assert.ok(!gotFoo, "foo hasn't been accessed during module loading");
+  assert.equal(program.useFoo(), foo, "foo mock works");
+  assert.ok(gotFoo, "foo has been accessed only when we first try to use it");
+};
```
Gozala: I think you may have mixed something up here. Your changes make access to built-in modules lazy that is passed as `modules` option to a loader. While here you test that `globals` are lazy instead. I imagine that's also why the tests are failing on travis.

lib/toolkit/loader.js
ochameau: Oops yes that was misleading... I put the const, hopefully this is clear enough now?

lib/toolkit/loader.js
ochameau: Oh right, I completely mixed up things, I original defined a setter in order to ensure that `exports` behavior is the same than this one defined over here:
https://github.com/ochameau/addon-sdk/blob/lazyModules/lib/toolkit/loader.js#L699-L700
```
 exports: { enumerable: true, writable: true, value: create(null),
configurable: true },
```
But here, we are past that, now exports is meant to be frozen on Module instance... right ?
I removed the setter.

test/test-loader.js
ochameau: You know, I still forget to do `git add`...!!

ochameau: But I fixed the test to actually check modules and not globals.

Gozala:
@ochameau code itself looks good, but we can take it as long as tests fail

screen shot 2015-02-17 at 14 37 17

With this fixed r+ from me.

PR #1848, opened 2015-01-28
Create unload tests for dev panel.

Gozala:


Gozala:
@erikvold can you please take a look at this ?

test/test-dev-panel.js
```diff
@@ -316,4 +316,97 @@ exports["test createView panel"] = test(function*(assert) {
...
+  // try catch since we don't care if it's exception or different
+  // panel and devtools may start throwing in a future.
+  try {
+    const toolbox = yield openToolbox(MyPanel);
+    const otherPanel = yield getCurrentPanel(toolbox);
+
+    assert.ok(!(otherPanel instanceof MyPanel),
+              "requested panel was not open as it was unregistered");
+  } catch (error) {
+    assert.pass("failed to open toolbox with panel that was unregistered");
```
erikvold: we should evaluate the error message here to make sure it is what we expect it to be.

test/test-dev-panel.js
```diff
@@ -316,4 +316,97 @@ exports["test createView panel"] = test(function*(assert) {
...
+  // At the moment devtools won't throw if requested panel isn't
+  // present and will just open different one, but we still use
+  // try catch since we don't care if it's exception or different
+  // panel and devtools may start throwing in a future.
+  try {
+    const toolbox = yield openToolbox(MyPanel);
+    const otherPanel = yield getCurrentPanel(toolbox);
+
+    assert.ok(!(otherPanel instanceof MyPanel),
+              "requested panel was not open as it was unregistered");
```
erikvold: shouldn't we `assert.fail()` if it gets this far?

test/test-dev-panel.js
```diff
@@ -316,4 +316,97 @@ exports["test createView panel"] = test(function*(assert) {
...
+            "current panel is not instance of MyPanel");
+
+  yield closeToolbox();
+
+  // At the moment devtools won't throw if requested panel isn't
+  // present and will just open different one, but we still use
+  // try catch since we don't care if it's exception or different
+  // panel and devtools may start throwing in a future.
+  try {
+    const toolbox = yield openToolbox(MyPanel);
```
erikvold: this should not throw right? if so can we move it out the the `try/catch`?

test/test-dev-panel.js
```diff
@@ -316,4 +316,97 @@ exports["test createView panel"] = test(function*(assert) {
...
```
erikvold: same as above.

test/test-dev-panel.js
```diff
@@ -316,4 +316,97 @@ exports["test createView panel"] = test(function*(assert) {
...
+    const otherPanel = yield getCurrentPanel(toolbox);
+
+    assert.ok(!(otherPanel instanceof MyPanel),
+              "requested panel was not open as it was unregistered");
+  } catch (error) {
+    assert.pass("failed to open toolbox with panel that was unregistered");
```
erikvold: same as above

andymckay:
Closing due to inactivity.

PR #1854, opened 2015-02-02
Bug 1128478 - sdk/panel's show/hide events not emitted if contentScriptWhen != 'ready'

Rob--W:
To verify whether the patch worked, I followed the following steps:

1. Update tests only and change `if (packaging.isNative) {` to `if (packaging.isNative && false) {` because otherwise all panel tests would be skipped.
2. Ran `JPM_FIREFOX_BINARY=/usr/bin/firefox-nightly node ./bin/jpm-test.js --type modules --filter test-panel:contentScriptWhen=start`
3. Observed that the new test failed.
4. Apply the fix.
5. Repeat step 2.
6. Observe that the new test passes.
7. Remove the `&& false` modification at step 1 and submit the patch.

This patch fixes a regression caused by https://bugzilla.mozilla.org/show_bug.cgi?id=1083391.

### DIFF
--- a/test/test-panel.js
+++ b/test/test-panel.js
@@ -1297,7 +1297,6 @@ exports["test panel addon global object"] = function*(assert) {
 exports["test panel load doesn't show"] = function*(assert) {
   let loader = Loader(module);
 
-  let showCount = 0;
   let panel = loader.require("sdk/panel").Panel({
     contentScript: "addEventListener('load', function(event) { self.postMessage('load'); });",
     contentScriptWhen: "start",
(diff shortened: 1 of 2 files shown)


test/test-panel.js
```diff
@@ -1335,6 +1334,28 @@ exports["test panel load doesn't show"] = function*(assert) {
...
+  });
+
+  let shown = false;
+  panel.once("show", function() {
+    shown = true;
+  });
+
+  panel.show();
+
+  yield wait(1000);
```
erikvold: replace this with:

    yield new Promise(resolve => {
      panel.once("show", resolve);
      panel.show();
    })

Rob--W: Done.

PR #1856, opened 2015-02-05
Bug 942511 - Fixes undefined properties of ActiveTab

colbat:


lib/sdk/tabs/observer.js
```diff
@@ -39,14 +39,28 @@ const Observer = Class({
...
     this.on("select", tab => {
       const selected = this[selectedTab];
+
+      // Bug 942511: this[selectedTab] is never updated when a popup window is
+      // closed. This leads to the situation where the properties of ActiveTab 
+      // are undefined because ActiveTab relates to the closed popup window. To 
+      // work around this situation we emit "activate" on the selected tab when
+      // this[selectedTab] was not updated (not equal to null) and when the 
+      // properties of ActiveTab are undefined.
+      const activeTabId = require("../tabs").activeTab.id;
```
erikvold: use this instead please:

    // at the top
    { getActiveTab, getTabs, getTabId } = require("./utils");

    // then use
    const activeTabId = getTabId(getActiveTab());

test/tabs/test-firefox-tabs.js
```diff
@@ -1057,6 +1059,41 @@ exports.testFaviconGetterDeprecation = function (assert, done) {
   });
 }
 
+// related to bug 942511
```
erikvold: nit adding another comment with a link to the bug here would be useful

test/tabs/test-firefox-tabs.js
```diff
@@ -1057,6 +1059,41 @@ exports.testFaviconGetterDeprecation = function (assert, done) {
...
+    tabs.open({
+      url: 'about:blank',
+      onReady: function (tab) {
+        tabID = tab.id;
+        tab.attach({
+          contentScript: 'var popup = window.open("about:blank");' +
+                         'popup.close();'
+        });
+
+        setTimeout(function() {
```
erikvold: we are trying to avoid using `setTimeout`s in our tests when possible, because they lead to intermittent failures and slower tests.  We can avoid using `setTimeout` in this case by either using another method to open the window, or by listening to the [observer service for window notifications](https://developer.mozilla.org/en-US/docs/Observer_Notifications#Windows).

lib/sdk/tabs/observer.js
```diff
@@ -39,14 +39,28 @@ const Observer = Class({
...
+      if (!activeTabId && selected) {
+        this[selectedTab] = tab;
+        emit(this, "activate", this[selectedTab]);
```
erikvold: won't this mean the `"activate"` event could emit twice on `this[selectedTab]` for each activation?

lib/sdk/tabs/observer.js
```diff
@@ -39,14 +39,28 @@ const Observer = Class({
...
       const selected = this[selectedTab];
+
+      // Bug 942511: this[selectedTab] is never updated when a popup window is
+      // closed. This leads to the situation where the properties of ActiveTab 
+      // are undefined because ActiveTab relates to the closed popup window. To 
+      // work around this situation we emit "activate" on the selected tab when
+      // this[selectedTab] was not updated (not equal to null) and when the 
+      // properties of ActiveTab are undefined.
+      const activeTabId = require("../tabs").activeTab.id;
+      if (!activeTabId && selected) {
```
erikvold: why don't we call `emit(this, 'deactivate', selected);` here?  if there is a good reason we should include this as a comment.

lib/sdk/tabs/observer.js
```diff
@@ -39,14 +39,28 @@ const Observer = Class({
     // that before emitting "activate" on selected tab.
     this.on("select", tab => {
       const selected = this[selectedTab];
+
+      // Bug 942511: this[selectedTab] is never updated when a popup window is
+      // closed. This leads to the situation where the properties of ActiveTab 
+      // are undefined because ActiveTab relates to the closed popup window. To
```
erikvold: what exactly is `ActiveTab` in this comment?  is it the `tab` variable? the result of `getActiveTab()`? the `selected` variable or something else?

test/tabs/test-firefox-tabs.js
```diff
@@ -1057,6 +1059,41 @@ exports.testFaviconGetterDeprecation = function (assert, done) {
...
+          let activeTabID = tabs.activeTab.id;
+          if (activeTabID) {
+              assert.equal(tabID, activeTabID, 'ActiveTab properties are correct');
+          } 
+          else {
+            assert.fail('ActiveTab properties undefined on popup closed');
+          }
+          setPref(OPEN_IN_NEW_WINDOW_PREF, defaultOpenWindowPref);
+          setPref(DISABLE_POPUP_PREF, defaultPopupPref);
```
erikvold: we should add before and after tests to this file to make sure that these prefs are in the state they should be.

For an example of using these see the [sidebar tests](https://github.com/colbat/addon-sdk/blob/activetab-properties-undefined-942511/test/test-ui-sidebar.js#L1545-L1552)

colbat:
Thank you for the review :) Regarding the bug itself I might have found another way to fix it, I think it's cleaner and less confusing. I removed what I did first and I just made a change in the if clause line 67, it appears to fix the bug. However I'm not sure about it because I'm not sure what was the purpose of the original if clause: `if (getActiveTab(chromeWindow) === this[selectedTab])` when the bug occurred it was never `true`: `getActiveTab(chromeWindow)` was referring to the tab of the closed popup window and `this[selectedTab]` was referring to the tab of the first window. This means that `this[selectedTab]` was never set to `null` when the popup window was closed, so the tab of the first window was never activated when the select event was handled.

Regarding the tests, actually I used a `setTimeout` because the bug occurred only when the activate event was fired again, so not just after the popup window was closed. And thank you for pointing out that to me, it was not the best way to do indeed. Instead I used the observer from "sdk/windows/observer" to listen to `activate`

colbat:
@erikvold Could you review this please?

The test passed on my machine, but the travis build failed. And I don't think it's related to the changes I made. From the travis log:
```
console.info: addon-sdk: executing './test/test-tabs.testOnLoadEventWithImage'
console.error: addon-sdk: 
  fail:
  Timed out (after: START)
console.error: addon-sdk: 
  fail:
  Should not be any unexpected tabs open
console.trace: addon-sdk: 
_ecated/unit-test.js 108 fail
_ecated/unit-test.js 338 done/<
_/Promise-backend.js 867 Handler.prototype.process
_/Promise-backend.js 746 this.PromiseWalker.walkerLoop
_/Promise-backend.js 688 this.PromiseWalker.scheduleWalkerLoop/<
```

benmccann:
@colbat thanks for the PR. Would you be able to fix the merge conflict?

colbat:
I fixed the merge conflict but it appears that the bug was actually fixed by 53d1725c7f31544ae5c700a30d26ff6be9559b4f. So I removed my changes since it's not needed anymore. But I think it's still useful to keep the unit test.

canuckistani:
@erikvold this looks like it applies cleanly - have you taken a look at it?

benmccann:
@colbat can you rebase against master?

colbat:
Sure, it's now rebased against master and I fixed the small merge conflicts

benmccann:
@erikvold @Mossop would you be able to review this? i've hit this bug a lot, so I think adding a test for it is a great idea

test/tabs/test-firefox-tabs.js
```diff
@@ -1191,10 +1194,63 @@ exports.testTabDestroy = function(assert, done) {
...
+            setPref(OPEN_IN_NEW_WINDOW_PREF, defaultOpenWindowPref);
+            setPref(DISABLE_POPUP_PREF, defaultPopupPref);
+            tab.close(done);
+          }
+        });
+      }
+    });
+};
+
+before(exports, (name, assert) => resetPopupPrefs);
```
Mossop: Do we really need to do this before each test? You're already doing it after each test so it seems redundant.

test/tabs/test-firefox-tabs.js
```diff
@@ -1191,10 +1194,63 @@ exports.testTabDestroy = function(assert, done) {
   })
 };
 
+// related to bug 942511
+// https://bugzilla.mozilla.org/show_bug.cgi?id=942511
+exports['test active tab properties defined on popup closed'] = function (assert, done) {
+    let defaultOpenWindowPref = getPref(OPEN_IN_NEW_WINDOW_PREF);
+    let defaultPopupPref = getPref(DISABLE_POPUP_PREF);
```
Mossop: Please use 2 space indent

test/tabs/test-firefox-tabs.js
```diff
@@ -1191,10 +1194,63 @@ exports.testTabDestroy = function(assert, done) {
...
+                         'popup.close();'
+        });
+        
+        windowObserver.once('close', () => {
+          popupClosed = true;
+        });
+
+        windowObserver.on('activate', () => {
+          // Only when the 'activate' event is fired after the popup was closed.
+          if(popupClosed) {
```
Mossop: Add a space after "if"

test/tabs/test-firefox-tabs.js
```diff
@@ -1191,10 +1194,63 @@ exports.testTabDestroy = function(assert, done) {
...
+            popupClosed = false;
+            let activeTabID = tabs.activeTab.id;
+            if (activeTabID) {
+                assert.equal(tabID, activeTabID, 'ActiveTab properties are correct');
+            }
+            else {
+              assert.fail('ActiveTab properties undefined on popup closed');
+            }
+            setPref(OPEN_IN_NEW_WINDOW_PREF, defaultOpenWindowPref);
+            setPref(DISABLE_POPUP_PREF, defaultPopupPref);
```
Mossop: You reset these in the after block anyway so no need to change them here

test/tabs/test-firefox-tabs.js
```diff
@@ -1191,10 +1194,63 @@ exports.testTabDestroy = function(assert, done) {
...
+          popupClosed = true;
+        });
+
+        windowObserver.on('activate', () => {
+          // Only when the 'activate' event is fired after the popup was closed.
+          if(popupClosed) {
+            popupClosed = false;
+            let activeTabID = tabs.activeTab.id;
+            if (activeTabID) {
+                assert.equal(tabID, activeTabID, 'ActiveTab properties are correct');
```
Mossop: Two space indent please.

lib/sdk/tabs/observer.js
```diff
@@ -42,12 +42,12 @@ const Observer = Class({
...
       if (selected !== tab) {
         if (selected) {
-          emit(this, 'deactivate', selected);
+          emit(this, "deactivate", selected);
         }
 
         if (tab) {
           this[selectedTab] = tab;
-          emit(this, 'activate', this[selectedTab]);
+          emit(this, "activate", this[selectedTab]);
```
Mossop: None of the changes in this file seem necessary anymore.

Mossop:
This looks good aside from one or two style nits. If you can update and ideally squash down to a single commit I can merge this.

colbat:
@Mossop Thank you for having taking the time to review this. I've did the changes and I've squashed everything down to a single commit. I've removed the preferences resets from the unit test itself to leave it only inside the after block, but I was wondering, is it the right thing to do or should I do the reverse? (removing it from the after block and resetting the preferences inside the unit test itself)

Mossop:
Thanks

PR #1857, opened 2015-02-07
Bug 839983 - Test that window.title is undefined when the window is closed

erikvold:


andymckay:
Bug is marked as resolved fixed, closing.

PR #1862, opened 2015-02-13
Bug 1129003: Make page-mod e10s safe and add e10s support modules

Mossop:
The first commit adds the remote/parent and remote/child e10s support modules that make it possible to load a module in the child processes (remoteRequire) and hooks up add-on specific event emitters for communication.
The second commit uses these to make page-mod e10s safe by moving the bulk of the module into the child process.

There is one main change here that probably shouldn't affect anyone but is unavoidable. When a page is hidden in history (frozen) its worker no longer throws exceptions when trying to send it messages, there is no way to know if a worker is frozen when the message is first sent. Instead we just queue the messages and they are delivered when the worker thaws.


lib/framescript/FrameScriptManager.jsm
```diff
@@ -3,33 +3,27 @@
...
 "use strict";
 
-const globalMM = Components.classes["@mozilla.org/globalmessagemanager;1"].
-                 getService(Components.interfaces.nsIMessageListenerManager);
+const { classes: Cc, interfaces: Ci } = Components;
+
+const globalMM = Cc["@mozilla.org/globalmessagemanager;1"].
+                 getService(Ci.nsIMessageListenerManager);
+
+const EXPORTED_SYMBOLS = ["getNewLoaderID"];
```
erikvold: we need to use `this.EXPORTED_SYMBOLS` in order to support b2g

lib/framescript/FrameScriptManager.jsm
```diff
@@ -3,33 +3,27 @@
...
+const PATH = __URI__.replace('framescript/FrameScriptManager.jsm', '');
 
-  loadedTabEvents = true;
-  globalMM.loadFrameScript(PATH + 'tab-events.js', true);
+// Builds a unique loader ID for this runtime. We prefix with the SDK path so
+// overriden versions of the SDK don't conflict
+let LOADER_ID = 0;
+function getNewLoaderID() {
+  return PATH + ":" + LOADER_ID++;
 }
```
erikvold: need a `this.getNewLoaderID = getNewLoaderID` here for b2g.

lib/framescript/content.jsm
```diff
@@ -0,0 +1,91 @@
...
+"use strict";
+
+const { utils: Cu, classes: Cc, interfaces: Ci } = Components;
+const { Loader } = Cu.import('resource://gre/modules/commonjs/toolkit/loader.js', {});
+const { Services } = Cu.import('resource://gre/modules/Services.jsm');
+
+const cpmm = Cc['@mozilla.org/childprocessmessagemanager;1'].
+             getService(Ci.nsISyncMessageSender);
+
+const EXPORTED_SYMBOLS = ["registerContentFrame"];
```
erikvold: `this.EXPORTED_SYMBOLS` for b2g

lib/framescript/content.jsm
```diff
@@ -0,0 +1,91 @@
...
+    unregisterContentFrame(contentFrame);
+  }, false);
+
+  frames.add(contentFrame);
+
+  for (let addon of addons.values()) {
+    if ("child" in addon)
+      addon.child.registerContentFrame(contentFrame);
+  }
+}
```
erikvold: `this.registerContentFrame = registerContentFrame` for b2g

lib/framescript/content.jsm
```diff
@@ -0,0 +1,91 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+"use strict";
+
+const { utils: Cu, classes: Cc, interfaces: Ci } = Components;
+const { Loader } = Cu.import('resource://gre/modules/commonjs/toolkit/loader.js', {});
```
erikvold: we should use `__URI__.replace(...)` here so that overloading continues working.

test/addons/chrome/install.rdf
```diff
@@ -0,0 +1,26 @@
...
+              <em:id>{ec8030f7-c20a-464f-9b0e-13a3a9e97384}</em:id>
+              <em:minVersion>33.0a1</em:minVersion>
+              <em:maxVersion>38.0</em:maxVersion>
+</Description>
+</em:targetApplication>
+
+
+    </Description>
+
+</RDF>
```
erikvold: this is file necessary?

test/addons/e10s-remote/main.js
```diff
@@ -0,0 +1,581 @@
...
+const REMOTE_MODULE = "./remote-module";
+
+const { Loader } = require('sdk/test/loader');
+const { getTabs, openTab, closeTab, setTabURL, getBrowserForTab, getURI } = require('sdk/tabs/utils');
+const { getMostRecentBrowserWindow } = require('sdk/window/utils');
+const { cleanUI } = require("sdk/test/utils");
+const { setTimeout } = require("sdk/timers");
+
+const { Cu } = require('chrome');
+const { async } = Cu.import('resource://gre/modules/Task.jsm', {}).Task;
```
erikvold: nit we can just require this `const { Task: async } = require('resource://gre/modules/Task.jsm');`

test/addons/e10s-remote/main.js
```diff
@@ -0,0 +1,581 @@
...
+    [remoteProcess] = yield processAttach;
+    assert.ok(remoteProcess.isRemote, "Process should be remote");
+    // And create a new frame in the remote process
+    [newFrame] = yield frameAttach;
+    assert.equal(newFrame.process, remoteProcess, "New frame should be in the remote process");
+
+    setTabURL(tab, "about:blank");
+
+    loader.unload();
+    yield cleanUI();
```
erikvold: let's just add one of these at the end of the file to reduce the repetition:

    after(exports, function*(name, assert) {
      yield cleanUI();
    });

test/addons/chrome/install.rdf
Mossop: Uhh no, in fact I have no idea where it or bootstrap.js came from as I didn't write them.

lib/sdk/content/page-mod.js
```diff
@@ -0,0 +1,249 @@
...
+const { frames, process } = require('../remote/child');
+const { Cc, Ci, Cu } = require('chrome');
+
+const pagemods = new Map();
+const styles = new WeakMap();
+let styleFor = (mod) => styles.get(mod);
+
+// Bind observer
+const { addObserver, removeObserver } = Cc['@mozilla.org/observer-service;1'].
+                                          getService(Ci.nsIObserverService);
```
erikvold: is there some reason for not using `sdk/system/events` here?

Mossop: Right, this is probably worthy of a comment and/or follow-up bug, but, until SDK add-ons are marked multiprocessCompatible registering for the document-element-inserted observer notification causes the add-on shims to forward those events from the child process. When it does so the subject parameter is a CPOW to the object in the child and any attempt to use it results in blocking cross-process messaging, part of what we're trying to avoid here. `sdk/system/events` triggers this when it does a bunch of checks on the subject to try to convert it back to an SDK object (https://github.com/mozilla/addon-sdk/blob/master/lib/sdk/system/events.js#L66). So I bypassed that by just using the raw observer service and then down in onContentWindow the first check is whether subject is a CPOW which costs nothing and we can bail out then.

It would be possible to use `sdk/core/observer` without adding extra CPOW traffic but I think that involves declaring a new class and creating an instance of it which seemed overkill.

We might be able to avoid the observer service entirely and use DOM events instead but I was having difficulty getting one or two tests passing in that case and I wanted to get this in before trying harder with that.

lib/sdk/tabs/utils.js
```diff
@@ -294,7 +294,9 @@ function getTabForBrowser(browser) {
         return tab;
     }
   }
-  return null;
+
+  let tabbrowser = browser.getTabBrowser && browser.getTabBrowser()
+  return tabbrowser && tabbrowser.getTabForBrowser(browser);
```
erikvold: this should all return `null` if `tabbrowser` is falsey

lib/sdk/remote/utils.js
```diff
@@ -0,0 +1,60 @@
...
+  let listener = (event, ...args) => {
+    emit(target, event, extraArg, ...args);
+  }
+
+  source.on('*', listener);
+
+  return function detach() {
+    source.off('*', listener);
+  }
+}
```
erikvold: let's move this to `sdk/event/utils` and add tests for it.

erikvold: I think you may be able to just modify `require("sdk/event/utils").pipe` to support this use case.

lib/sdk/remote/utils.js
```diff
@@ -0,0 +1,60 @@
...
+const AttachPoint = Class({
+  implements: [ List ],
+
+  attachItem: function(item) {
+    addListItem(this, item);
+
+    let detachPort = forwardEvents(item.port, this.port, item);
+    let detachMain = forwardEvents(item, this, item);
+
+    item.on('detach', () => {
```
erikvold: `item.on('detach', () => {` => `item.once('detach', () => {`

lib/sdk/content/mod.js
```diff
@@ -56,6 +56,10 @@ function detach(modification, target) {
   else {
     let documents = iterator(modification);
     for (let document of documents) {
+      let window = document.defaultView;
+      // The window might have already gone away
+      if (!window)
+        continue;
```
erikvold: Just realized we have no tests for this file :( so I made [bug 1134839](https://bugzilla.mozilla.org/show_bug.cgi?id=1134839) about that.

erikvold: I think we should handle this is a separate bug.

lib/sdk/remote/child.js
```diff
@@ -0,0 +1,265 @@
...
+exports.process = process;
+
+process.port.emit = (...args) => {
+  mm.sendAsyncMessage('sdk/remote/process/message', {
+    loaderID,
+    args
+  });
+}
+
+function processMessageReceived({ data }) {
```
erikvold: nit please add a comment here about what this conditional below is for, like `// ignore messages from other loaders/add-ons`

lib/sdk/remote/parent.js
```diff
@@ -0,0 +1,313 @@
...
+
+    processMap.set(this.id, this);
+    processes.attachItem(this);
+  },
+
+  dispose: function() {
+    emit(this, 'detach');
+    processMap.delete(this.id);
+    ns(this).messageManager.removeMessageListener('sdk/remote/process/message', ns(this).messageReceived);
+    ns(this).messageManager.removeMessageListener('child-process-shutdown', this.destroy);
```
erikvold: `ns(this).messageManager = null;`

lib/sdk/remote/parent.js
```diff
@@ -0,0 +1,313 @@
...
+        args
+      });
+    };
+
+    frameMap.set(ns(this).messageManager, this);
+  },
+
+  dispose: function() {
+    emit(this, 'detach');
+    ns(this).messageManager.removeMessageListener('sdk/remote/frame/message', ns(this).messageReceived);
```
erikvold: `ns(this).messageManager = null;`

lib/sdk/remote/parent.js
```diff
@@ -0,0 +1,313 @@
...
+const ns = Namespace();
+
+let processMap = new Map();
+
+function processMessageReceived({ target, data }) {
+  if (data.loaderID != loaderID)
+    return;
+  emit(this.port, ...data.args);
+}
+
```
erikvold: nit can we add some description about what a `Process` is here please, something like `// Process is a child process which contains one to many frames/tabs`.

lib/sdk/remote/utils.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  source.on('*', listener);
+
+  return function detach() {
+    source.off('*', listener);
+  }
+}
+
+const AttachPoint = Class({
```
erikvold: nit some comments on what `AttachPoint` is and does would be useful here, I don't think it's obvious.

test/addons/e10s-remote/main.js
```diff
@@ -0,0 +1,581 @@
...
+// Test that we find the right number of processes and that messaging between
+// them works and none of the streams cross
+exports["test process list"] = function*(assert) {
+  let loader = new Loader(module);
+  let { processes, frames, remoteRequire } = loader.require('sdk/remote/parent');
+
+  let processCount = 0;
+  processes.forEvery(processes => {
+    processCount++;
+  });
```
erikvold: this can just be `processes.forEvery(() => processCount++);`

test/addons/e10s-remote/main.js
```diff
@@ -0,0 +1,581 @@
...
+  yield waitForProcesses(loader);
+
+  let remoteProcesses = Array.filter(processes, process => process.isRemote);
+  let localProcesses = Array.filter(processes, process => !process.isRemote);
+
+  assert.equal(localProcesses.length, 1, "Should always be one process");
+
+  if (isE10S) {
+    assert.equal(remoteProcesses.length, 1, "Should be one remote processes");
+  } else {
```
erikvold: nit newline here

test/addons/e10s-remote/main.js
```diff
@@ -0,0 +1,581 @@
...
+    assert.equal(remoteProcesses.length, 0, "Should be no remote processes");
+  }
+
+  assert.equal(processCount, processes.length, "Should have seen all processes");
+
+  processCount = 0;
+  processes.forEvery(process => {
+    processCount++;
+  });
```
erikvold: `processes.forEvery(() => processCount++);`

test/addons/e10s-remote/main.js
```diff
@@ -0,0 +1,581 @@
...
+          resolve();
+        }
+      });
+    }
+    processes.on('attach', listener);
+  });
+
+  console.log("Remote ready");
+  return { processes, frames, remoteRequire };
+});
```
erikvold: Can we please move the above helper functions to another file?

test/addons/e10s-remote/main.js
```diff
@@ -0,0 +1,581 @@
...
+
+if (isE10S) {
+  console.log("Testing in E10S mode");
+  // We expect a child process to already be present, make sure that is the case
+  mainWindow.XULBrowserWindow.forceInitialBrowserRemote();
+
+  // Check that we see a process stop and start
+  exports["test process restart"] = function*(assert) {
```
erikvold: just have this function return early if `isE10S` is true, to avoid indenting the whole test.

test/addons/e10s-remote/main.js
```diff
@@ -0,0 +1,581 @@
...
+    let window = getMostRecentBrowserWindow();
+
+    let tabs = getTabs(window);
+    assert.equal(tabs.length, 1, "Should have just the one tab to start with");
+    let tab = tabs[0];
+
+    let loader = new Loader(module);
+    let { processes, frames, remoteRequire } = yield waitForProcesses(loader);
```
erikvold: `remoteRequire` doesn't appear to be used in this test.

test/addons/e10s-remote/main.js
```diff
@@ -0,0 +1,581 @@
...
+}
+else {
+  console.log("Testing in non-E10S mode");
+}
+
+// Test that we find the right number of processes and that messaging between
+// them works and none of the streams cross
+exports["test process list"] = function*(assert) {
+  let loader = new Loader(module);
+  let { processes, frames, remoteRequire } = loader.require('sdk/remote/parent');
```
erikvold: `remoteRequire` doesn't appear to be used in this test.

test/addons/remote/main.js
```diff
@@ -0,0 +1,581 @@
...
+  closeTab(tab);
+  loader2.unload();
+
+  yield cleanUI();
+}
+
+// Check that the child frame has the right properties
+exports["test frame properties"] = function*(assert) {
+  let loader = new Loader(module);
+  let { processes, frames, remoteRequire } = yield waitForProcesses(loader);
```
erikvold: `remoteRequire` doesn't appear to be used in this test.

test/addons/remote/main.js
```diff
@@ -0,0 +1,581 @@
...
+
+  let promise = new Promise(resolve => {
```
erikvold: `let promise = new Promise(resolve => {` => `yield new Promise(resolve => {`

erikvold: `let promise = new Promise(resolve => {` => `yield new Promise(resolve => {`

lib/sdk/system/runtime.js
```diff
@@ -15,6 +15,7 @@ exports.inSafeMode = runtime.inSafeMode;
 exports.OS = runtime.OS;
 exports.processType = runtime.processType;
 exports.widgetToolkit = runtime.widgetToolkit;
+exports.processID = runtime.processID;
```
erikvold: We should add tests for this.

erikvold:
We should add a test that the `WorkerChild` instances are destroyed when unload occurs, or when `worker.destroy()` is called in the main process.

Same for `ChildPageMod` and anything else like it.

test/addons/remote/main.js
```diff
@@ -0,0 +1,581 @@
...
+  loader2.unload();
+
+  yield cleanUI();
+}
+
+// Test that unloading unregisters frames DOM events
+exports["test unload removes frames event listeners"] = function*(assert) {
+  let window = getMostRecentBrowserWindow();
+  let loader = new Loader(module);
+  let { processes, frames, remoteRequire } = yield waitForProcesses(loader);
```
erikvold: `remoteRequire` doesn't appear to be used here.

test/addons/e10s-remote/main.js
```diff
@@ -0,0 +1,581 @@
...
+  yield promiseEvent(process2.port, 'sdk/test/pong');
+
+  loader2.unload();
+};
+
+// Test that unloading the loader unloads the child instances
+exports["test unload"] = function*(assert) {
+  let window = getMostRecentBrowserWindow();
+  let loader = new Loader(module);
+  let { processes, frames, remoteRequire } = yield waitForProcesses(loader);
```
erikvold: `remoteRequire` doesn't appear to be used here.

test/addons/e10s-remote/main.js
```diff
@@ -0,0 +1,581 @@
...
+  closeTab(tab);
+
+  yield cleanUI();
+}
+
+// Test that unloading the loader causes the child to see detach events
+exports["test detach on unload"] = function*(assert) {
+  let window = getMostRecentBrowserWindow();
+  let loader = new Loader(module);
+  let { processes, frames, remoteRequire } = yield waitForProcesses(loader);
```
erikvold: `remoteRequire` doesn't appear to be used here.

test/addons/e10s-remote/main.js
```diff
@@ -0,0 +1,581 @@
...
+  closeTab(tab);
+
+  yield cleanUI();
+}
+
+// Test that DOM event listener on the frame object works
+exports["test frame event listeners"] = function*(assert) {
+  let window = getMostRecentBrowserWindow();
+  let loader = new Loader(module);
+  let { processes, frames, remoteRequire } = yield waitForProcesses(loader);
```
erikvold: `remoteRequire` doesn't appear to be used here.

test/addons/remote/main.js
```diff
@@ -0,0 +1,581 @@
...
+  let hash = getURI(tab).replace(/.*#/, "");
+  assert.equal(hash, "unloaded:shutdown", "Saw the correct hash change.")
+
+  closeTab(tab);
+
+  yield cleanUI();
+}
+
+// Test that unloading the loader causes the child to see detach events
+exports["test detach on unload"] = function*(assert) {
```
erikvold: `exports["test detach on unload"]` to `exports["test frame detach on unload"]`

lib/sdk/remote/utils.js
Mossop: So the code here does the distinction that events emitted to the process/frame objects (or their ports) don't include the process/frame as the first argument whereas the events emitted the the processes/frames objects do. Having since seen that tabs/windows doesn't do this, all the events received the relevant target as the first argument, I think it makes more sense to do that. It's consistent and at least a few times while developing this I made a mistake because of it so I've switched to that which makes pipe work just fine (though I'm a little worried by GC cycles being an issue)

lib/sdk/remote/utils.js
Mossop: The name bothered me a little too so I've renamed to EventParent for now. Open to other suggestions too.

lib/sdk/content/mod.js
Mossop: Well I made this change because a test fails without it.

PR #1867, opened 2015-02-20
Insert addon-sdk tag after prompt prefix

dcolascione:
This change tries to make the addon SDK insert the (addon-sdk) tag in the right place when PS1 begins with terminal control codes and whitespace.


erikvold:
Sorry, we aren't going to be releasing any new versions of cfx in the future, so the code patched here is deprecated.

PR #1868, opened 2015-02-21
s/substract/hyphenminus/

canuckistani:
I can't map to the 'hyphenminus' key in current nightly - this 1-line fix seems to correct it. See https://bugzilla.mozilla.org/show_bug.cgi?id=1135312


canuckistani:
@erikvold can you review this?

erikvold:
@canuckistani we should add a test for this

canuckistani:
@erikvold tests added!

andymckay:
Not merging based on the last comment, but closing due to inactivity since then.

PR #1876, opened 2015-02-26
Bug 971098 Implementing add-on level module overrides

erikvold:


erikvold:
Not sure about the `"sdk"` key, what do you think @gozala @jsantell ?

jsantell:
Is this to namespace to prevent future collisions with others using package.json? If so, we should probably make it more specific like "moz-sdk" or "jetpack" or something

erikvold:
@jsantell y, maybe "jetpack" is better

lib/toolkit/loader.js
```diff
@@ -623,17 +625,26 @@ const Require = iced(function Require(loader, requirer) {
...
     // TODO should get native Firefox modules before doing node-style lookups
     // to save on loading time
     if (isNative) {
       // If a requireMap is available from `generateMap`, use that to
       // immediately resolve the node-style mapping.
+      // TODO: write more tests for this use case
       if (requireMap && requireMap[requirer.id])
         requirement = requireMap[requirer.id][id];
 
+      let { overrides } = manifest.jetpack;
```
jsantell: won't this break if `jetpack` is undefined?

jsantell: nm I see you add it always if not listed, but then check below if it exists (which it should always exist, right?)

lib/toolkit/loader.js
```diff
@@ -623,17 +625,26 @@ const Require = iced(function Require(loader, requirer) {
...
+      if (overrides) {
+        for (let key in overrides) {
+          if (id == key || (id.substr(0, key.length + 1) == (key + "/"))) {
```
jsantell: Is this statement checking for trailing slashes? Should add a comment, a lot of this require code is hard to read (partially my fault) for lack of comments and scenarios

test/fixtures/native-overrides-test/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "name": "native-overrides-test",
+  "main": "index.js",
+  "dependencies": {
+    "foo": "*",
+    "fs-extra": "*"
+  },
+  "jetpack": {
+    "overrides": {
+      "fs": "sdk/io/fs",
+      "overload": "foo"
```
jsantell: Does this cover the scenario where we have `"sdk/tabs": "./tabs"`, where you're using your own version of "fs" or possibly an sdk built-in, or just want global access to a module? I think that was the complexity part I ran into before, but if this supports that scenario, then we should test it, otherwise, have a separate bug for it

lib/toolkit/loader.js
```diff
@@ -764,6 +778,16 @@ function Loader(options) {
     sharedGlobalBlacklist: ["sdk/indexed-db"]
   }, options);
 
+  // Create overrides defaults, none at the moment
```
jsantell: Hopefully default node mappings later :D :D :D

PR #1887, opened 2015-03-06
Bug 1140361 - Remove __noSuchMethod__ from the Add-on SDK

jandem:
Two commits:

(1) Don't use __noSuchMethod__ in context-menu/test-helper.js
(2) Fix the 'fake' console objects to match the real console object. We removed __noSuchMethod__ from that one in bug 1139794.


jandem:
Landed this on m-c, as that's what I'm used to :)

PR #1889, opened 2015-03-10
Bug 1132753 and bug 1141275: Fix l10n/html to avoid CPOWs and run tests on jpm

Mossop:
Split into two commits, one fixes bug 1141275 and enabled the l10n test in jpm, the second fixes the CPOW problem. See the commit messages for more details.


erikvold:
Pretty sure this is done, please reopen if I am making a mistake.

PR #1891, opened 2015-03-11
Bug 1142186: Only delete the test preferences.

Mossop:
This test was deleting the entire test add-ons preferences, this includes the preferences that the mochitest-jetpack harness uses to override the values in sdk/self which was breaking logging testing in later tests.


test/test-native-options.js
```diff
@@ -163,7 +163,11 @@ exports.testSimplePrefs = function(assert) {
   assert.strictEqual(simple.prefs.test3, "1", "test3 is '1'");
   assert.strictEqual(simple.prefs.test4, "red", "test4 is 'red'");
 
-  branch.deleteBranch('');
+  for (let setting of parent.children) {
```
erikvold: nit please add a comment here

PR #1901, opened 2015-03-17
Bug 1142314 - Output the exact build of Firefox being testing

erikvold:


Mossop:
We already output some information about the running build here: https://github.com/mozilla/addon-sdk/blob/master/lib/sdk/test/harness.js#L599. It probably makes more sense to just add the build ID to that.

PR #1904, opened 2015-03-17
Bug 1135219: Avoid opening channels for obviously bad URIs. Don't resolve node_module paths above the loader's rootURI.

Mossop:


lib/toolkit/loader.js
```diff
@@ -431,9 +444,14 @@ const nodeResolve = iced(function nodeResolve(id, requirer, { rootURI }) {
   if ((resolvedPath = loadAsDirectory(fullId)))
     return stripBase(rootURI, resolvedPath);
 
+  // If the requirer is an absolute URI then the node module resolution below
+  // won't work correctly as we prefix everything with rootURI
+  if (isAbsoluteURI(requirer))
+    return void 0;
```
erikvold: It doesn't appear that we have any tests for this use case in test-native-loader.js, we should add some since I'm assuming we are using the use case somehow?

lib/toolkit/loader.js
```diff
@@ -420,6 +429,10 @@ const nodeResolve = iced(function nodeResolve(id, requirer, { rootURI }) {
   // Resolve again
   id = Loader.resolve(id, requirer);
 
+  // If this is already an absolute URI then there is no resolution to do
+  if (isAbsoluteURI(id))
+    return void 0;
```
erikvold: We should add tests too test-native-loader.js for this use case as well.

lib/toolkit/loader.js
Mossop: It's a weird case and it's possible it only appears in tests, but at the very least test-addon-bootstrap triggers this. The way it loads the code here https://github.com/mozilla/addon-sdk/blob/master/test/test-addon-bootstrap.js#L37 means we end up hitting this case with the require calls from `sdk/addon/bootstrap` but only when it requires non-relative paths.

PR #1906, opened 2015-03-18
Bug 1144742: Clear promise debugging observer after tests and clear all added DOM event listeners on unload.

Mossop:
Made sdk/event/dom remove various event listeners even when the window for those elements have been closed. In theory this shouldn't be necessary but in practice this is causing some reference cycle between the DOM elements and SDK code. The simplest way to break it is just to remove the listeners when the window closes.

Also clears our promise debugging observer which was keeping the test harness and so a lot of stuff inside that loader alive.


Mossop:
Landed as 6b627247edf7340843a455bbb319ca73461c05e2

PR #1911, opened 2015-03-21
 	Bug 1135219 Adding a test for logging assertions from the toolkit/loader

erikvold:


andymckay:
Bug 1135219 is marked as fixed in bugzilla.

PR #1914, opened 2015-03-23
Reimplement tabs and windows APIs.

Mossop:
This reimplements the Firefox tabs and windows API using heritage and other more standard APIs instead of traits. It isn't intended to fix any bugs in particular (though I'm pretty sure the behaviour is much more sane when there is no open browser window), merely to re-implement what we have in a way that we can iterate on more easily in the future.

it doesn't merge the Firefox and Fennec implementations though it does bring them closer together. I'm not sure that that is a useful goal at this point since it looks like there are a bunch of timing differences that would complicate maintaining them both together. Instead I've tried to make sure we use tabs/utils wherever possible which we can keep platform agnostic.

All of the existing tests pass with just two changes. tab.destroy is no longer a thing in Firefox. It has never been in Fennec and tests show that we only care that it doesn't do much so I don't think that's a controversial change. The `this` variable in event listeners passed to browserWindows.open has changed to be the opened window rather than browserWindows. That kind of fell out of the implementation and is fixable but it seemed to make more sense to me.

Almost everything is e10s-compliant now. Setting tab titles has a minor change to support this, noted in the code. getThumbnail is already disabled in the e10s case. tab.attach still needs work (https://bugzilla.mozilla.org/show_bug.cgi?id=1146603). tab.readyState is not e10s compatible and it isn't clear what we should do here (https://bugzilla.mozilla.org/show_bug.cgi?id=1146606)


Mossop:
One additional test fix for testBrowserWindowCreationOnActivate was to wait for the window to focus before checking that tabs saw an activate event. The old code would send activate events for new tabs whenever a new window opened regardless of whether the window eventually received focus. In probably almost all cases it will get focused but the new code waits for an explicit focus event on the window before sending the tab activate event. This can happen before or after the window load completes.

lib/sdk/windows/tabs-fennec.js
```diff
@@ -9,7 +9,7 @@ const { browserWindows } = require('./fennec');
 const { windowNS } = require('../window/namespace');
 const { tabsNS, tabNS } = require('../tabs/namespace');
 const { openTab, getTabs, getSelectedTab, getTabForBrowser: getRawTabForBrowser,
-        getTabContentWindow } = require('../tabs/utils');
+        getTabContentWindow, isTab } = require('../tabs/utils');
```
erikvold: nit this doesn't appear to be used.

test/windows/test-firefox-windows.js
```diff
@@ -29,7 +29,7 @@ exports.testOpenAndCloseWindow = function(assert, done) {
   browserWindows.open({
     url: "data:text/html;charset=utf-8,<title>" + title + "</title>",
     onOpen: function(window) {
-      assert.equal(this, browserWindows, "The 'this' object is the windows object.");
+      assert.equal(this, window, "The 'this' object is the window object.");
```
erikvold: I don't think we should change this, some add-on may depend on it.

test/test-tab.js
```diff
@@ -127,12 +127,6 @@ exports["test behavior on close"] = function(assert, done) {
                      "After being closed, tab attributes are undefined (url)");
         assert.equal(tab.index, undefined,
                      "After being closed, tab attributes are undefined (index)");
-        if (app.is("Firefox")) {
-          // Ensure that we can call destroy multiple times without throwing;
-          // Fennec doesn't use this internal utility
-          tab.destroy();
-          tab.destroy();
-        }
```
erikvold: why is this removed?

lib/sdk/tabs.js
```diff
@@ -20,14 +20,3 @@ else {
 }
 
 const tabs = module.exports;
-
-// Implement `modelFor` function for the Tab instances.
-// Finds a right model by iterating over all tab models
-// and finding one that wraps given `view`.
-modelFor.when(isTab, view => {
```
erikvold: nit remove the require statement above for `viewFor` and `modelFor`

test/tabs/test-firefox-tabs.js
```diff
@@ -1079,8 +1079,8 @@ exports.testNoDeadObjects = function(assert, done) {
           // 3. Arrange to close the tab once the second page loads.
           tab.on("load", function() {
             tab.close(function() {
-              var TABS = loader.sandbox("sdk/tabs/tab-firefox").TABS;
-              assert.equal(TABS.length, 0, "no dead objects");
+              var modelsFor = loader.sandbox("sdk/tabs/tab-firefox").modelsFor;
```
erikvold: `loader.sandbox("sdk/tabs/tab-firefox").modelsFor` => `loader.require("sdk/model/core").modelFor`

test/tabs/test-firefox-tabs.js
```diff
@@ -1079,8 +1079,8 @@ exports.testNoDeadObjects = function(assert, done) {
...
+              assert.equal(modelsFor.get(tab), undefined, "didn't retain the closed tab");
```
erikvold: this will always be true since the `tab` instance comes from a different loader instance.

lib/sdk/tabs/observer.js
```diff
@@ -87,6 +88,11 @@ const Observer = Class({
     for (let chromeWindow of browserWindowIterator()) {
       this.observe(chromeWindow);
     }
+
+    when(_ => {
+      // Don't dispatch a deactivate event during unload.
+      this[selectedTab] = null;
```
erikvold: do we have a test for this?

lib/sdk/tabs/tabs-firefox.js
```diff
@@ -1,68 +1,120 @@
...
+    let activeDomWin = getMostRecentBrowserWindow();
+    if (!activeDomWin)
+      return null;
+    return modelFor(getSelectedTab(activeDomWin));
+  },
+
+  open: function(options) {
+    options = Options(options);
 
+    let windows = require('../windows').browserWindows;
```
erikvold: please add the ` // TODO: BUG 792670 - remove dependency below` comment here

lib/sdk/tabs/tab-firefox.js
```diff
@@ -3,309 +3,307 @@
...
+  get window() {
+    if (isClosed(this))
       return undefined;
-    if (this._tab.getAttribute('remote')) {
-      console.error('This method is not supported with E10S');
-      return BLANK;
-    }
-    return getThumbnailURIForWindow(this._contentWindow);
+
+    require('../windows');
```
erikvold: please add the `// TODO: Remove this dependency` comment here

lib/sdk/tabs/tab-firefox.js
```diff
@@ -3,309 +3,307 @@
...
+    let tabElement = viewsFor.get(this);
+    let domWindow = tabElement.ownerDocument.defaultView;
+    return modelFor(domWindow);
+  },
+
+  get readyState() {
+    // Not e10s compliant
```
erikvold: nit please make this a `TODO` so that we can more easily find it

lib/sdk/tabs/tab-firefox.js
```diff
@@ -3,309 +3,307 @@
...
+  close: function(callback) {
+    let tabElement = viewsFor.get(this);
+
+    if (!tabElement || !tabElement.parentNode) {
+      callback();
       return;
-    // BUG 792946 https://bugzilla.mozilla.org/show_bug.cgi?id=792946
-    // TODO: fix this circular dependency
-    let { Worker } = require('./worker');
-    return Worker(options, this._contentWindow);
```
erikvold: I guess we can close [bug 792946](https://bugzilla.mozilla.org/show_bug.cgi?id=792946) now.

erikvold: oh nvm I see it below.

lib/sdk/tabs/tab-firefox.js
```diff
@@ -3,309 +3,307 @@
...
       return;
+
+    activateTab(viewsFor.get(this));
+  }),
+
+  getThumbnail: function() {
+    if (isClosed(this))
+      return BLANK;
+
+    // Not e10s compliant
```
erikvold: nit `// Not e10s compliant` => `// TODO: Make this e10s compliant`

lib/sdk/windows/firefox.js
```diff
@@ -1,356 +1,225 @@
...
+const { emit, setListeners } = require('../event/core');
+const { once } = require('../dom/events');
+const { EventTarget } = require('../event/target');
+const { getSelectedTab } = require('../tabs/utils');
+const { Cc, Ci } = require('chrome');
+const { Options } = require('../tabs/common');
+const system = require('../system/events');
+const { ignoreWindow, isPrivate, isWindowPBSupported } = require('sdk/private-browsing/utils');
+const { isPrivateBrowsingSupported } = require('sdk/self');
+const { data } = require('../self');
```
erikvold: nit combine the lines above

lib/sdk/windows/firefox.js
```diff
@@ -1,356 +1,225 @@
...
-        tab.inNewWindow = true;
+const { modelFor } = require('../model/core');
+const { emit, setListeners } = require('../event/core');
+const { once } = require('../dom/events');
+const { EventTarget } = require('../event/target');
+const { getSelectedTab } = require('../tabs/utils');
+const { Cc, Ci } = require('chrome');
+const { Options } = require('../tabs/common');
+const system = require('../system/events');
+const { ignoreWindow, isPrivate, isWindowPBSupported } = require('sdk/private-browsing/utils');
```
erikvold: nit `require('sdk/private-browsing/utils')` => `require('../private-browsing/utils')`.

test/test-tab.js
Mossop: As far as I can see from the existing implementation (there are no docs), calling destroy on a tab effectively hides it from the tabs APIs. It will no longer be visible in the tabs list nor will any events be fired for it. I couldn't think of a single reason for doing that and it seems like a footgun to me. We also have no testing of it other than this test that we can call destroy. I could try to reimplement the behaviour but since the Fennec API doesn't implement this either I think it makes more sense to drop it. I could add a no-op method too I guess.

erikvold:
It appears this patch removes all of our uses of `sdk/deprecated/list` so let's remove that here as well please.

test/tabs/test-firefox-tabs.js
Mossop: Worse than that, I should be using viewFor not modelFor!

lib/sdk/tabs/tab-firefox.js
Mossop: I don't know that it is possible to remove this dependency or why we would need to. This function returns an object from the windows API so it has to be loaded.

lib/sdk/tabs/observer.js
Mossop: test-firefox-tabs:testAutomaticDestroyEventDeactivate covers this.

Mossop:
fced8ac6de92ba5f8c5948bb22213ddd0ba41991 adds tests for the this property in all event listeners I could think of. They pass on both old and new code with one exception. In the old code event listeners on the `tabs` object actually get the internal object from `sdk/windows/tabs-firefox` which in the prototype chain of `tabs` because of the hack at https://github.com/mozilla/addon-sdk/blob/master/lib/sdk/tabs/tabs-firefox.js#L62. That internal object is gone now but the new code just uses the `tabs` object which shouldn't break anything except crazy code.

The windows event listeners are a little crazy in that the open and close events always have `browserWindows` as this regardless of how you add them. The only straightforward way to emulate this was to implement an `emitOnObject` function in `sdk/events/core`. The old code uses the deprecated _emitOnObject function at https://github.com/mozilla/addon-sdk/blob/master/lib/sdk/deprecated/events.js#L142.

test/test-tab.js
erikvold: I'm not sure what the fennec comment is about.  The reason this test was added as because calling `tab.destroy();` previously didn't clean itself up properly and it was only discovered because someone pointed out calling destroy on a tab twice was producing an error.  Alex fixed this one, and we determined that part of the test should include calling destroy more than once, to ensure that does not produce an error.  So I think we should keep this, I don't see any reason to change it yet anyhow, I don't understand how it could be a footgun.

lib/sdk/tabs/tab-firefox.js
erikvold: It creates a circular dependency, that is why we can't have it at the top of the document with all of the other `require` statements.  We need to break the circular dependency.  Probably the easiest way to do this is to define a module which exports both the `Window` constructor and a `Tab` constructor.

erikvold:
Thanks for making that change in fced8ac, we should probably change this at some point I'm not sure if Irakli really wanted them this way, but I think it's best to figure that out in another bug and then we can be more organized in letting people know about the change.

benmccann:
@Mossop thanks for this! FYI, the PR needs a rebase

Mossop:
I did some further investigation of how tab.destroy() behaves in the existing code, since it isn't documented anywhere. After calling tab.destroy the tab still exists in the tabs and window.tabs lists. No events for that tab are ever fired again. Functions and properties for that tab stop working, in effect it is a dead object. I've reimplemented all of this.

The one behaviour I found that I haven't yet implemented is how activeTab behaves with destroyed tabs. Here is what you can do with the current code:
```javascript
tab1.activate();
// now tabs.activeTab == tab1;
tab1.destroy();
// now tabs.activeTab == tab1;
tab2.activate();
// now tabs.activeTab == tab2;
// activate tab1 through the UI or another loader
// now tabs.activeTab == tab2 still
```
The new code just uses modelFor to get the activeTab, this is more reliable than tracking the activeTab manually IMO but it would make it hard to replicate this behaviour of returning the last selected tab if the currently selected tab is destroyed. It doesn't seem like a good behaviour to me either so I wanted to see what you'd like to do before I spent the time to implement it. My opinion is that we should just leave it returning the dead tab, seems more correct to me.

I'll rebase and squash this to master when I land, otherwise I think github will throw away a lot of the comments :(

lib/sdk/tabs/tab-firefox.js
```diff
@@ -3,309 +3,328 @@
...
-  // If called asked to return only existing wrapper,
-  // we should return null here as no matching Tab object has been found
-  if (existingOnly)
+  catch (e) {
     return null;
+  }
+}
+
+function tabEmit(tab, event, ...args) {
+  const allTabs = require('../tabs');
```
erikvold: I didn't notice this before, it looks like a new circular dependency which we can avoid by emitting on a new module, which the tabs module would then subscribe to.

lib/sdk/tabs/tab-firefox.js
```diff
@@ -3,309 +3,328 @@
...
+  let domWindow = tabElement.ownerDocument.defaultView;
+
+  if (ignoreWindow(domWindow))
+    return;
+
+  // Don't send events for tabs that are already closing
+  if (event != "close" && (tabElement.closing || !tabElement.parentNode))
+    return;
+
+  const allTabs = require('../tabs');
```
erikvold: same as above, we should remove this circular dependency.

lib/sdk/tabs/tab-firefox.js
```diff
@@ -3,309 +3,328 @@
...
+
+  get isPinned() {
+    return isDestroyed(this) ? undefined : viewsFor.get(this).pinned;
+  },
+
+  get window() {
+    if (isClosed(this))
+      return undefined;
+
+    require('../windows');
```
erikvold: please add a `// TODO: Remove the dependency on the windows module: bug 792670` comment here too

lib/sdk/tabs/tab-firefox.js
Mossop: So this isn't what I'd consider to be a circular dependency, requiring tabs does automatically require tab-firefox but that doesn't automatically require tabs. Can you explain to me why this delayed require of tabs is a problem because I really don't get it?

erikvold: If someone now makes a patch using `tabEmit` during the loading of this module, and we don't have a test which requires this module without having required the tabs modules previously (iirc), then this use case will break without our knowledge.  Basically the circular dependency makes the module much harder to maintain because it is more error prone in a complex way, it will require more tests as a result (ie loading and using "sdk/tabs/tab-firefox" in a loader instance where "sdk/tabs" is not previously loaded) now and in the future.

Mossop:
Updated again, ready for review.

erikvold:
Closed by https://github.com/mozilla/addon-sdk/commit/53d1725c7f31544ae5c700a30d26ff6be9559b4f

PR #1916, opened 2015-03-25
Bug 1146545 Adding gulp commands to clean or apply a patch from m-c

erikvold:


erikvold:
Mentioned the wrong bug number.. should be bug 1147245

PR #1920, opened 2015-03-25
Bug 1119601 Adding a gulp command to automatically generate the test/addons/jetpack-addon.ini file

erikvold:


bin/node-scripts/update-ini.js
```diff
@@ -0,0 +1,50 @@
...
+    var result = {};
+
+    fs.readdir(addonsDir, function(err, files) {
+      var folders = files.filter(function(file) {
+        return fs.statSync(path.resolve(addonsDir, file)).isDirectory();
+      }).sort();
+
+      folders.forEach(function(folder) {
+        var oldData = data[folder + ".xpi"];
+        if (oldData) {
```
Mossop: Looks like this if check is unnecessary

PR #1921, opened 2015-03-26
Bug 1147747 - Fix isArray fallback function and isArgument to return correct value.

arai-a:
https://bugzilla.mozilla.org/show_bug.cgi?id=1147747

### DIFF
--- a/lib/sdk/lang/type.js
+++ b/lib/sdk/lang/type.js
@@ -125,9 +125,7 @@ exports.isGenerator = isGenerator;
  *    isArray([1, 2, 3])  // true
  *    isArray({ 0: 'foo', length: 1 }) // false
  */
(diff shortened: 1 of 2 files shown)


lib/sdk/lang/type.js
```diff
@@ -126,7 +126,7 @@ exports.isGenerator = isGenerator;
  *    isArray({ 0: 'foo', length: 1 }) // false
  */
 var isArray = Array.isArray || function isArray(value) {
-  Object.prototype.toString.call(value) === "[object Array]";
+  return Object.prototype.toString.call(value) === "[object Array]";
```
erikvold: this code won't ship in a version of Firefox that does not have `Array.isArray` now, so we can just remove the funciton after `Array.isArray || `

PR #1922, opened 2015-03-26
Bug 1140759 - Fix in tree consumers that use non-standard Function.prototype.isGenerator in Add-on SDK.

arai-a:
https://bugzilla.mozilla.org/show_bug.cgi?id=1140759

### DIFF
--- a/test/test-addon-bootstrap.js
+++ b/test/test-addon-bootstrap.js
@@ -21,7 +21,7 @@ const BOOTSTRAP_REASONS = {
   ADDON_DOWNGRADE : 8
 };
 
-exports["test install/startup/shutdown/uninstall all return a promise"] = function(assert) {
+exports["test install/startup/shutdown/uninstall all return a promise"] = function*(assert) {
   let uri = require.resolve("./fixtures/addon/bootstrap.js");
   let id = "test-min-boot@jetpack";
   let bootstrapScope = create({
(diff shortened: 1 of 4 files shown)


test/test-addon-bootstrap.js
```diff
@@ -21,7 +21,7 @@ const BOOTSTRAP_REASONS = {
   ADDON_DOWNGRADE : 8
 };
 
-exports["test install/startup/shutdown/uninstall all return a promise"] = function(assert) {
+exports["test install/startup/shutdown/uninstall all return a promise"] = function*(assert) {
```
erikvold: this change seems unnecessary

test/test-event-core.js
```diff
@@ -261,7 +261,7 @@ exports['test listen to all events'] = function(assert) {
     'wildcard listener called for unbound event name');
 };
 
-exports['test once'] = function(assert, done) {
+exports['test once'] = function*(assert, done) {
```
erikvold: this change seems unnecessary, and it would have to remove the `done` argument.

test/test-event-dom.js
```diff
@@ -15,7 +15,7 @@ function countListeners(target, type) {
   return listeners.filter(listener => listener.type == type).length;
 }
 
-exports['test window close clears listeners'] = function(assert) {
+exports['test window close clears listeners'] = function*(assert) {
```
erikvold: this change seems unnecessary, and it would have to remove the `done` argument.

arai-a: maybe I should've commented here, sorry for long delay.
Those are required because of the change in Part 2.
Those functions are passed to isGenerator function, so all legacy generator in tests should be replaced with ES6 generator.

PR #1950, opened 2015-04-27
Bug 850865 - Stop hijacking console in addon documents

erikvold:


lib/sdk/content/sandbox.js
```diff
@@ -198,8 +197,6 @@ const WorkerSandbox = Class({
     // WorkerSandbox so we can communicate with content script
     model.emitToContent = result;
 
-    let console = new PlainTextConsole(null, getInnerId(window));
-
     // Handle messages send by this script:
     setListeners(this, console);
```
ochameau: You have to also modify `setListeners`.

PR #1951, opened 2015-04-29
Bug 1159554 - Implement some commit message tags which will filter travis tests

erikvold:
...s tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 language: node_js
 node_js:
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -1,7 +1,7 @@
 sudo: false
 language: node_js
 node_js:
-  - "0.10"
+  - "0.12"
```
jsantell: I'd add both, 0.10 is still way more popular and used.

PR #1952, opened 2015-04-29
Bug 1159580 - Use travis cache to avoid redoing spell checks

erikvold:


erikvold:
before path every travis build was running ~2secs to do spell checks, and used 2 network requests which intermittently fail (rarely though).  This reduces the time to 11ms on average, makes the tests more reliable, and only tests changes to the docs.

PR #1953, opened 2015-04-29
Bug 1147741 Part 1 - Add travis tests that the jetpack-addon.ini file is up to date

erikvold:


bin/node-scripts/test.ini.js
```diff
@@ -0,0 +1,39 @@
...
+        throw err;
+      }
+      var text = data.toString();
+      var expected = "";
+
+      ini.makeAddonIniContent()
+      .then(function(contents) {
+        expected = contents;
+
+        setTimeout(function end() {
```
Mossop: Why the setTimeout here?

erikvold: I need to look into this, but promises may be swallowing mocha test failures, at least that seemed to be the case while I was working on this.

PR #1960, opened 2015-05-05
Bug 1161493 - responseURL get property for response object in sdk/request

skipme:
Allow get modified URL in case of a redirection. Forwarded responseURL property of encapsulated xhr.

### DIFF
--- a/lib/sdk/request.js
+++ b/lib/sdk/request.js
@@ -164,8 +164,9 @@ const Response = Class({
   initialize: function initialize(request) {
     response(this).request = request;
   },
+  // more about responseURL: https://bugzilla.mozilla.org/show_bug.cgi?id=998076
+  get url() response(this).request.responseURL,
   get text() response(this).request.responseText,
-  get responseURL() response(this).request.responseURL,
   get xml() {
     throw new Error("Sorry, the 'xml' property is no longer available. " +
                     "see bug 611042 for more information.");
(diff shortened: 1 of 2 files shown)


test/test-request.js
```diff
@@ -92,6 +92,35 @@ exports.testStatus200 = function (assert, done) {
...
+
+  var req = Request({
+    url: "http://localhost:" + port + "/" + basename,
+    onComplete: function (response) {
+      assert.equal(this, req, "`this` should be request");
+      assert.equal(response.status, 200);
+      assert.equal(response.statusText, "OK");
+      assert.equal(response.headers["Content-Type"], "text/plain");
+      assert.equal(response.text, movedContent);
+      assert.equal(response.responseURL, movedUrl);
```
erikvold: this should be strict equal, to make sure `response.responseURL` is a string, or test that separately.

lib/sdk/request.js
```diff
@@ -165,6 +165,7 @@ const Response = Class({
     response(this).request = request;
   },
   get text() response(this).request.responseText,
+  get responseURL() response(this).request.responseURL,
```
erikvold: The other property names here appear to have removed the `response` prefix, so in order to follow that pattern this line should be:

    get url() response(this).request.responseURL

skipme:
Thank you, looks better.

erikvold:
Thanks!

PR #1961, opened 2015-05-05
Support Add-on SDK env detection for Add-ons

DecipherCode:
Adds a new global constant for environment detection.

Alternative method is to append the following to sdk/system/globals.
https://github.com/mozilla/addon-sdk/blob/master/lib/sdk/system/globals.js#L22
```js
Object.defineProperty(exports, 'JETPACK', {
 __proto__: null,
 value: {}
});
```


erikvold:
One can use the [`engine-details`](https://www.npmjs.com/package/engine-details) like so:

    require("engine-details").is("jetpack");

https://github.com/erikvold/engine-details/blob/master/index.js#L9-L13

I'd rather not add things that can be done by other means.

DecipherCode:
By other redundant means, of course.

See: https://bugzilla.mozilla.org/show_bug.cgi?id=1161376

PR #1962, opened 2015-05-06
Fixed python detection.

kkapsner:
Python detection is not working in Win 7 because the standard software
folder may contain spaces: "C:\Program Files\..."


erikvold:
We aren't accepting new commits for the python code, we're trying to remove it, please use jpm instead.  https://blog.mozilla.org/addons/2015/02/26/jpm-replaces-cfx-for-firefox-38/

erikvold:
Including styling in the package.json is also something that we do not want to do.

kkapsner:
OK - I will try jpm.

Regarding the styling: I understand your reasons not to include styling in the package.json, but there should be some way to style the preference pane. This was just the simplest way...

I would really like to be able to include a CSS-file into the preference pane.

PR #1963, opened 2015-05-06
Avoid travis errors about no output in 10mins

erikvold:


bin/node-scripts/utils.js
```diff
@@ -49,6 +49,23 @@ exports.spawn = spawn;
 function run (cmd, options, p) {
   return new Promise(function(resolve) {
     var output = [];
+    var lastOutputLimit = 0;
+    var mins = 0;
+    var minsInvterval = setInterval(function() {
```
Mossop: s/minsInvterval/minsInterval/

PR #1972, opened 2015-05-14
Fixed Bug with plural rules: if json.language() returns null getRulesForLocale will produce a warning but will return a valid function. Therefore getRulesForLocale ("en") is never called.

kkapsner:
I know you will not include the python fix...


erikvold:
I'm sorry we can't take the python related commits here.

kkapsner:
I know - but the plural rule bug is still valid...

PR #1973, opened 2015-05-17
Fix invalid method syntax used in URL constructor

evilpie:
This blocks landing a patch in bug 1158222, where I want to disallow this syntax. (It's against the ES6 spec)


evilpie:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions
https://people.mozilla.org/~jorendorff/es6-draft.html#sec-method-definitions

Note the required { }

ZER0:
LGTM, r+

PR #2011, opened 2015-06-04
Bug 1163408 avoid using setImmediate and use the same delay that we must wait for

erikvold:


test/test-ui-action-button.js
```diff
@@ -716,7 +717,11 @@ exports['test button tab state'] = function*(assert) {
     'tab badgeColor inherited from global');
 
   // check the node properties
-  yield wait();
+  yield new Promise(resolve => {
+    let target = {};
+    once(target, "ready", resolve);
+    emit(target, "ready");
+  });
```
Mossop: I'm not really sure what this is waiting for. Why is waiting for an async promise to resolve any better than setImmediate?

erikvold: Well I'm not sure tht it is better, but it seems like using setImmediate is failing us.

PR #2018, opened 2015-06-11
Stop parenthesizing array/object pattern destructuring assignment targets

jswalden:
These syntaxes are invalid in ES6:

    var v;
    ([v]) = [0]; // BAD way to assign v = 0
    ({v}) = { v: 0 }; // BAD way to assign v = 0

ES6 says that when destructuring assignment is used (or when a variable or variables are declared using a destructuring pattern), array/object patterns must not be parenthesized.

There's no reason at all to parenthesize array patterns (you can assign directly to an unparenthesized pattern), so this only rarely happens.  But some people have parenthesized object patterns, as a way to disambiguate them from normal JS block statements.  However, ES6 doesn't sanction such disambiguation attempts.  The right way to fix this is to parenthesize the entire assignment, not just the object pattern.

There are a few uses of this in the addon SDK, that need fixing before I can land my patch for https://bugzilla.mozilla.org/show_bug.cgi?id=1146136 upstream.  Mind pulling this in for me?  (And, is it a problem if my patch upstream also patches the imported addon-sdk/ files, for you to overwrite next time you pull?  Or must I wait for the next Github -> mozilla-central import to happen first?)

### DIFF
--- a/lib/sdk/io/buffer.js
+++ b/lib/sdk/io/buffer.js
@@ -247,11 +247,11 @@ Object.defineProperties(Buffer.prototype, {
     value: function(string, offset, length, encoding = 'utf8') {
       // write(string, encoding);
       if (typeof(offset) === 'string' && Number.isNaN(parseInt(offset))) {
-        ([offset, length, encoding]) = [0, null, offset];
+        [offset, length, encoding] = [0, null, offset];
       }
       // write(string, offset, encoding);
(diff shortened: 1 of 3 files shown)


canuckistani:
Hey @jswalden - this looks great and I understand it is blocking landing of bug 1146136 - I've pinged @jsantell about this and we are going to see about getting this into m-c, hopefully soon.

PR #2020, opened 2015-07-07
Bug 922956 - Add unit test to prevent regressions

jaragunde:
The bug seems to be fixed, I'm adding a unit test to prevent regressions and close the bug for good.

### DIFF
--- a/test/tabs/test-firefox-tabs.js
+++ b/test/tabs/test-firefox-tabs.js
@@ -1234,6 +1234,33 @@ exports['test active tab properties defined on popup closed'] = function (assert
   });
 };
 
+// related to bug 922956
+// https://bugzilla.mozilla.org/show_bug.cgi?id=922956
+exports["test ready event after window.open"] = function (assert, done) {
+  setPref(OPEN_IN_NEW_WINDOW_PREF, 2);
+  setPref(DISABLE_POPUP_PREF, false);
+
+  let firstRun = true;
+  tabs.on('ready', function onReady(tab) {
+    if (firstRun) {
+      assert.pass("tab ready callback after 1st window.open");
+      firstRun = false;
+      tab.close();
+    }
+    else {
+      assert.pass("tab ready callback after 2nd window.open");
+      tabs.removeListener('ready', onReady);
+      tab.close(done);
+    }
+  });
+
+  tabs.activeTab.attach({
+    contentScript: "window.open('about:blank');" +
+                   "window.open('about:blank', '', " +
+                   "'width=800,height=600,resizable=no,status=no,location=no');"
+  });
+}
+
 after(exports, function*(name, assert) {
   resetPopupPrefs();
   yield cleanUI();


benmccann:
@erikrose @Mossop @jsantell @Gozala would one of you guys be able to take a look at this to see if bug 922956 is really fixed?

benmccann:
ping. would really love it if someone could take a look at this one

erikrose:
I think you got the wrong Erik; you probably meant one of the 2 CCd on the ticket.

benmccann:
Sorry, yep, I meant to tag @erikvold

jaragunde:
I've slightly modified the test to include a call to window.open with some parameters to have that case covered.

test/test-window-events.js
```diff
@@ -60,4 +64,20 @@ exports["test browser events"] = function(assert, done) {
   let window = open();
 };
 
+// related to bug 922956
+// https://bugzilla.mozilla.org/show_bug.cgi?id=922956
+exports["test ready event after window.open"] = function (assert, done) {
+  setPref(OPEN_IN_NEW_WINDOW_PREF, 2);
+  setPref(DISABLE_POPUP_PREF, false);
```
Mossop: You need to reset these prefs at the end of the test or it may affect later tests.

Mossop: That may be what is causing the travis tests to fail, if not it looks like they warrant investigation.

jaragunde:
@Mossop Thanks for the review; I've added your suggestions and submitted a new patch. Unfortunately it did not pass Travis either; this time there is a weird error with no list of failed tests.

I'll keep looking into it.

Mossop:
Looks like some timeout is happening or something. The first failure, mozRequestAnimationFrame is due to a change in Firefox, I've fixed that in master so you might try rebasing and pushing again to see if that fixes the others too.

benmccann:
@jaragunde can you rebase and push again? Thanks!

benmccann:
@Mossop looks like this PR was updated if you want to take another look

test/tabs/test-firefox-tabs.js
```diff
@@ -1234,6 +1234,33 @@ exports['test active tab properties defined on popup closed'] = function (assert
...
+
+  let firstRun = true;
+  tabs.on('ready', function onReady(tab) {
+    if (firstRun) {
+      assert.pass("tab ready callback after 1st window.open");
+      firstRun = false;
+      tab.close();
+    }
+    else {
+      assert.pass("tab ready callback after 2nd window.open");
```
benmccann: wouldn't the test pass even if neither of the `assert.pass` lines were ever called?

how does this line get called? won't the test exit on the first call to `assert.pass`? (i'm not sure exactly what testing framework is being used here, but `assert.pass` will end a test as passing in most frameworks i've seen)

test/tabs/test-firefox-tabs.js
```diff
@@ -1234,6 +1234,33 @@ exports['test active tab properties defined on popup closed'] = function (assert
...
+      tabs.removeListener('ready', onReady);
+      tab.close(done);
```
jaragunde: @benmccann The test will fail if it does not end with a call to done(). That call happens in the line above, after tab is closed.

test/tabs/test-firefox-tabs.js
jaragunde: @benmccann In this case, assert.pass() just logs a message.

jaragunde:
Travis has failed again, but all the tests pass successfully on my local machine. Besides, the same set of tests are failing in another [unrelated pull request](https://travis-ci.org/mozilla/addon-sdk/builds/74027446), so I suspect the problem may not be related with this patch.

Mossop:
Yeah I ran try builds against the full Firefox tree yesterday and they came up green so something is broken with travis right now. No need to hold this up any longer, sorry for the caution there.

PR #2021, opened 2015-07-11
First attempt for hotkey simple-preference

brettz9:
Functional but hackish attempt to solve https://bugzilla.mozilla.org/show_bug.cgi?id=1123097 .

With package.json like:

```js
...
"preferences": [{
        "name": "opendialog",
        "title": "Open dialog",
        "type": "hotkey",
        "value": "ctrl-o"
    }]
```
...the user's "Open dialog" preference will show with "ctrl-o".

Similarly, lib/main could retrieve this as follows:

```js
exports.main = function () {
    var prefs = require('sdk/simple-prefs');
    console.log("Let's see the last saved value: " + prefs.prefs['opendialog']);
};
```

Anyways, this is just my first attempt, and I should warn that my health has made me fickle in following through with past attempts, but I am having a few good days, so wanted to try this out.

I didn't make any attempt to bake in metadata in the package.json preferences to invoke a particular extension function, and I think it is quite easy and flexible as is to just get the current value from preferences and use that. But my implementation no doubt needs some work. Would be grateful for a lookover.

### DIFF
--- a/lib/sdk/preferences/native-options.js
+++ b/lib/sdk/preferences/native-options.js
@@ -17,7 +17,7 @@ const { defer } = require("sdk/core/promise");
 const DEFAULT_OPTIONS_URL = 'data:text/xml,<placeholder/>';
 
 const VALID_PREF_TYPES = ['bool', 'boolint', 'integer', 'string', 'color',
-                          'file', 'directory', 'control', 'menulist', 'radio'];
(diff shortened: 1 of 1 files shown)


andymckay:
I'm closing this because it's incomplete at this time. I've followed up on bug 1123097 to see if we can find someone to mentor and review this bug.

PR #2022, opened 2015-07-13
Bug 1178609 - `join` in loader.js does not work with resource:///

ZER0:


jlongster:
Is the error in TravisCI related to this? I can't really tell.

ZER0:
> Is the error in TravisCI related to this? I can't really tell.

No, the cause of failure here is something I fixed on bug 1184643 (pull 2024).

PR #2025, opened 2015-07-21
Remove one useless __iterator__ use

evilpie:
We really need to move people using the addon-sdk away from using `__iterator__`, i.e. using for-in to iterate over collections etc. (Bug 1098617)

At least here somebody probably made a mistake and used for-of instead of for-in inside the if.


andymckay:
There are patches in bugzilla, I think we can close this PR.

PR #2029, opened 2015-08-07
Bug 1180955 - Add the ability to override the require function in loaders

ZER0:


ZER0:
I believe that Travis here is failing due the missing preference for the signed add-on; so it's unrelated. I filed bug 1192166, however, to keep track of this.

PR #2031, opened 2015-08-16
Bug 825016 - On Fennec, pass the id of the currently selected tab when opening a new tab in order to preserve back key navigation

buttercookie42:


andymckay:
Would this need a test?

buttercookie42:
Probably not a bad idea, but how/where are those tests actually run on Android? For the try server, Treeherder is showing the jetpack mochitests as available only for the desktop builds.

buttercookie42:
Continuing this [on Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=825016).

PR #2032, opened 2015-09-03
Bug 1103385 - Make json.language() defaults to "en". r=jsantell

fred-wang:


andymckay:
This has been merged and the bugzilla bug closed.

PR #2034, opened 2015-09-28
Bug 1208898 - missing sdk/io/file.open append mode fix

erjoalgo:
fix https://bugzilla.mozilla.org/show_bug.cgi?id=1208898

### DIFF
--- a/lib/sdk/io/file.js
+++ b/lib/sdk/io/file.js
@@ -129,20 +129,30 @@ exports.list = function list(path) {
 };
 
 exports.open = function open(filename, mode) {
-  var file = MozFile(filename);
-  if (typeof(mode) !== "string")
+  const file = MozFile(filename);
+  if (typeof (mode) !== "string") {
     mode = "";
+  }
 
   // File opened for write only.
-  if (/w/.test(mode)) {
-    if (file.exists())
+  if (/[aw]/.test(mode)) {
+    const exists = file.exists();
+    if (exists) {
       ensureFile(file);
-    var stream = Cc['@mozilla.org/network/file-output-stream;1'].
-                 createInstance(Ci.nsIFileOutputStream);
-    var openFlags = OPEN_FLAGS.WRONLY |
-                    OPEN_FLAGS.CREATE_FILE |
-                    OPEN_FLAGS.TRUNCATE;
-    var permFlags = parseInt("0644", 8); // u+rw go+r
+    }
+    const stream = Cc['@mozilla.org/network/file-output-stream;1'].
+	  createInstance(Ci.nsIFileOutputStream);
+    let openFlags = OPEN_FLAGS.WRONLY;
+
+    if (exists && /a/.test(mode)) {
+      openFlags |= OPEN_FLAGS.APPEND;
+    }
+    else {
+      openFlags |= OPEN_FLAGS.TRUNCATE |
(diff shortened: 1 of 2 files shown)


lib/sdk/io/file.js
```diff
@@ -129,20 +129,30 @@ exports.list = function list(path) {
...
+    const stream = Cc['@mozilla.org/network/file-output-stream;1'].
+	  createInstance(Ci.nsIFileOutputStream);
+    let openFlags = OPEN_FLAGS.WRONLY;
+
+    if (exists && /a/.test(mode)) {
+      openFlags |= OPEN_FLAGS.APPEND;
+    }
+    else {
+      openFlags |= OPEN_FLAGS.TRUNCATE |
+        OPEN_FLAGS.CREATE_FILE;
```
Mossop: You can include CREATE_FILE regardless of whether the file exists or not. If the file already exists it has no effect.

lib/sdk/io/file.js
```diff
@@ -150,23 +160,24 @@ exports.open = function open(filename, mode) {
...
+	  createInstance(Ci.nsIFileInputStream);
+    try {
+      stream.init(file, OPEN_FLAGS.RDONLY, 0, 0);
+    }
+    catch (err) {
+      throw friendlyError(err, filename);
+    }
+    return /b/.test(mode) ?
+      new byteStreams.ByteReader(stream) :
+      new textStreams.TextReader(stream);
```
Mossop: I can't see any reason to move all this into an else block.

lib/sdk/io/file.js
```diff
@@ -129,20 +129,30 @@ exports.list = function list(path) {
 };
 
 exports.open = function open(filename, mode) {
-  var file = MozFile(filename);
-  if (typeof(mode) !== "string")
+  const file = MozFile(filename);
+  if (typeof (mode) !== "string") {
     mode = "";
+  }
```
Mossop: It's true this existing code doesn't match the style guide perfectly but I'd rather leave it alone than lose blame at this point.

test/test-file.js
```diff
@@ -139,6 +139,36 @@ exports.testOpenNonexistentForWrite = function (assert) {
...
+  const expectedContents = stringA + stringB;
+  const actualContents = read(filename);
+  assert.equal(expectedContents, actualContents,
+    "should have appended to file instead of truncating");
+
+  file.remove(filename);
+  assert.ok(!file.exists(filename),
+    "file.exists() should return false after file.remove()");
+};
+
```
Mossop: Please test that appending to a missing file and an empty file works as expected

andymckay:
Closing due to lack of activity from the original contributor. If anyone else would like to fix this bug, please commit a patch to bug 1208898 in bugzilla.

erjoalgo:
I tried very hard to follow the style guides in this commit (there is a previous PR for the same issue, which was closed due to not following this style) and I'm disappointed that something as simple as this was closed now based on my actually following the style guide. It feels as if you're less interested in fixing bugs and improving the project and more interested in making a display of your authority over the project. If you are really this obsessed with a trailing bracket or an else statement, you should agree upon or provide an auto-formatting tool like jsfmt to deal with the issue sanely, because nobody cares to study several pages and documents of your particular style preferences.

andymckay:
There are least three comments on this patch that are unrelated to the style.

PR #2036, opened 2015-09-29
Bug 1209470 - Remove use of expression closure from Add-on SDK.

arai-a:
https://bugzilla.mozilla.org/show_bug.cgi?id=1209470

Green on try run with applying same fix to m-c (some difference in toplevel var vs let)
  https://treeherder.mozilla.org/#/jobs?repo=try&revision=a804d1db825f


Mossop:
Code changes should land in mozilla-central now.

PR #2041, opened 2015-11-13
Fixes bug 1224561: allow to load Browserify modules

emersion:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1224561

### DIFF
--- a/lib/toolkit/loader.js
+++ b/lib/toolkit/loader.js
@@ -978,7 +978,7 @@ Loader.generateMap = generateMap;
 // since node allows 'lib/index.js' without relative `./`
 function getManifestMain (manifest) {
   let main = manifest.main || './index.js';
-  if (typeof manifest.browser == 'string') {
+  if (typeof manifest.browser === 'string') {
     main = manifest.browser;
   }
   return isRelative(main) ? main : './' + main;
--- a/lib/toolkit/loader.js
+++ b/lib/toolkit/loader.js
@@ -978,6 +978,9 @@ Loader.generateMap = generateMap;
 // since node allows 'lib/index.js' without relative `./`
 function getManifestMain (manifest) {
   let main = manifest.main || './index.js';
+  if (typeof manifest.browser == 'string') {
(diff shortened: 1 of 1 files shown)


emersion:
@jsantell @Mossop @KWierso Could someone review this PR please?

emersion:
Hey please @tschneidereit @rpl @mikeconley @KWierso, this is a bitesize PR!

rpl:
@emersion Unfortunately I'm not sure that this change is enough to achieve what you want:
even if the loader will choose the `"browser"` entry-point for the npm modules, some of them will not work correctly because they will expect `global` to be a reference to a `window` object, which is not the case in an Addon SDK module (which run in a JS sandbox).

This Github issue is going to be closed, mainly because this kind of changes to the Addon SDK loader are definitely something that have to be carefully discussed and tested, and any issues and changes to the Addon SDK have to go through Bugzilla and mozilla-central repo from now on, and so it is better that the discussion to happen in the Bugzilla issue already opened.

In the mean time, I agree that sometimes being able to include a browserified module in an addon can be pretty useful and I'd like to provide you (and anyone which needs to use the browserified version of an npm module in an addon) a couple of alternative solutions:

- a WebExtensions addon can load a browserified bundle in any WebExtension page (e.g. the background page or a popup page) like it can be done in a webpage, because these WebExtensions pages are like webpages running in a tab (plus the extended APIs)

- an Addon SDK addon can use the `"sdk/page-worker"` module to create an hidden addon window which can load a browserified bundle, and once the hidden page is ready, the addon modules communicates with the page worker using message  exchange.

here is two basic example addons which I put together to sketch the approach:

- https://github.com/rpl/mozilla-addons-and-browserify

in the Addon SDK addon example the "messages exchange" is wrapped inside a promise-based API to provide a nicer API to the other addon modules,
it is not as simple as `require("crypto-browserify").randomBytes`, but it fully works on any of the current Firefox releases and it only uses High Level Addon SDK modules.

PR #2042, opened 2015-11-20
Bug 1226754 - Enable base64-encoding of long strings

Rob--W:


Rob--W:
@ZER0 Please review.

andymckay:
@Rob--W could you submit a patch in Bugzilla please for bug 1226754?

PR #2043, opened 2015-12-14
test on recent versions of node.js

nikolas:


andymckay:
Tests don't run in travis anymore, but in try-server in mozilla central. Thanks for the pull request, but we'll close this.

PR #2044, opened 2016-01-03
Bug 956461 - context menuitems have no id in the dom.

brettz9:
Add code for adding "id" attribute on context-menu items; fix test labeling for accesskey tests.

When attempting to run `npm test` locally, it hangs, so currently cannot confirm this PR is working properly.

### DIFF
--- a/lib/sdk/context-menu.js
+++ b/lib/sdk/context-menu.js
@@ -256,6 +256,17 @@ var labelledItemRules =  mix(baseItemRules, {
     ok: v => !!v,
     msg: "The item must have a non-empty string label."
   },
+  id: {
+    map: stringOrNull,
+    is: ["string", "undefined", "null"],
+    ok: v => {
+      if (!v) {
+        return true;
+      }
+      return typeof v == "string";
+    },
+    msg: "The item must have a non-empty string id"
+  },
   accesskey: {
     map: stringOrNull,
     is: ["string", "undefined", "null"],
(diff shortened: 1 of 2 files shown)


rpl:
@brettz9 The workflow for addon-sdk patches has recently changes and patches are now applied on a mozilla-central branch and submitted on bugzilla as hg patches.

I've converted your patch into hg and attached it to the original bugzilla issue:
https://bugzilla.mozilla.org/show_bug.cgi?id=956461

From an mozilla-central working tree you can run your test-case file using the following command lines:

```
## for non-e10s test run
mozilla-central$ ./mach mochitest addon-sdk/source/test/test-context-menu.js

## or for e10s enabled test-run
mozilla-central$ ./mach mochitest addon-sdk/source/test/test-context-menu.js --e10s 
```

brettz9:
Thank you for the info, @rpl , but it is kind of intimidating to have to go through all these hoops of using a different VCS, etc. when I'm only submitting a bug for a few of the users of my free add-on (actually add-ons) who are having problems because the SDK doesn't support ID's as needed by the Menu Wizard or such menu editing add-ons, and thus the Firefox-recommended way causes problems for these users. I hope someone can take a look at what I've tried to add and apply what should still be a simple fix for the users benefit.

andymckay:
We will close this pull request, the bug exists in mozilla central with the patch, follow ups can continue there.

PR #2045, opened 2016-02-05
Bug 1246107 - A spell error of delimeters in mozrunner/__init__.py

ccwenlin:
Just a spell error, see https://bugzilla.mozilla.org/show_bug.cgi?id=1246107


andymckay:
We don't think this is used any more, so thank you, but we'll close this.

PR #2046, opened 2016-03-08
bug 1238064 update README to remove github

andymckay:


andymckay:
Oops, should be a mozilla-central commit.

PR #2052, opened 2016-09-09
Deprecate the standalone SDK repository

palant:
As discussed on IRC, it should be more obvious that this repository is outdated.


wagnerand:
@andym would you mind taking this, please? I don't know who else has access.

andymckay:
Are you basically just deleting the entire repo? Because the patch is so large, I can't see the 5 lines that you've added. 

I wouldn't mind getting a ping from @Mossop or @ZER0, @Gozala and others who put the repo here before this gets merged.

For context, for a while now we've stopped doing the github - mozilla-central syncing and just been asking for bugzilla bugs and mercurial patches. I think this patch would make it obvious that this repo is no longer used and provide an end point so when people land here there is something to see rather than just deleting the repository and creating a 404.

palant:
You can see the end result under https://github.com/palant/addon-sdk/tree/deprecate-repo - everything is removed, merely a few lines in the README remain.

wagnerand:
@andymckay the file that has the five added lines is the readme file.

